PR #10, opened 2022-07-21
Merkle Proof Verification

### LINKED ISSUE #6: Merkle Proof Verification Library
#### Description

Merkle Proofs use a tree-based structure with leaves and a root. Given a leaf, root, and proof, the library should have a function that will verify the validity of the given proof. 

#### Motivation

Merkle Proofs are commonly used in airdrops. This library will be directly integrated into the sway-applications airdrop contract.

bitzoic:
## Type of change

- New Library

## Changes

- Added Merkle proof library
     - Process Merkle Proof Function
     - Verify Merkle Proof Function
     - Leaf Digest Function
     - Node Digest Function
     - Key Length function
- Added Merkle proof tests

## Notes

- There is commented out code until https://github.com/FuelLabs/fuels-rs/issues/353 is resolved
- The current use of Fuel-Merkle fails due to padding of `u8` on Sway side. Issue to track is here: https://github.com/FuelLabs/sway/issues/2594

## Related Issues

Closes #6


bitzoic:
CI is failing due to an update in `fuel-core`

Braqzen (commented):
I'm not familiar with merkle logic so here's comments instead

1. merkles_proof.sw line 43 uses `itterator` instead of `iterator`. Is this intentional or a typo?
2. Line 120 only returns the hash but there is a comment stating that a check is meant to occur?
3. I would move the data structures to the top (e.g. ProofError) and have the functions at the bottom
4. Does the forc version specified in the Cargo.toml do anything? I thought it was meant to use/default to the installed version that the user has
5. There is an async test called `test_merkle_proof_instance` which doesn't test anything and instead it creates and returns the instance. I think this should be renamed to be reflective of its behaviour
6. Instead of commenting tests out I think it's better to `#[ignore]` them
7. wen formatter v2

bitzoic:
> 4. Does the forc version specified in the Cargo.toml do anything? I thought it was meant to use/default to the installed version that the user has

This should be needed for CI as I am running `cargo test` once in the `tests/src/` directory instead of `forc test` in every `tests/src/test_projects/*` directory. There is no Forc.toml file here so `forc test` will not work.

> 7. wen formatter v2

Agreed.

simonr0204:
Do you have a link to the spec for the merkle tree implementation this proof verification conforms to?

sway_libs/src/merkle_proof.sw
```diff
@@ -0,0 +1,151 @@
...
+
+fn hash_pair(a: b256, b: b256) -> b256 {
+    if a <= b {
+        // Hash(a + b)
+        sha256((a, b))
+    } else {
+        // Hash(b + a)
+        sha256((b, a))
+    }
+}
```
adlerjohn, line 20: This looks like you're trying to copy Uniswap's Merkle tree. This library should instead conform to https://github.com/FuelLabs/fuel-merkle-sol, which implements [RFC-6962](https://datatracker.ietf.org/doc/html/rfc6962) Merkle tree proof verification.

sway_libs/src/merkle_proof.sw
```diff
@@ -0,0 +1,151 @@
...
+    if total_hashes > 0 {
+        hashes.get(total_hashes - 1).unwrap()
+    } else if merkle_leaves_len > 0 {
+        merkle_leaves[0]
+    } else {
+        proof
+    }
+}
+
+/// This function will compute a merkle root given a leaf and corresponding proof.
```
matt-user, line 92:
```suggestion
/// This function will compute and return a merkle root given a leaf and corresponding proof.
```

sway_libs/src/merkle_proof.sw
```diff
@@ -0,0 +1,151 @@
...
+        let a = if leaf_pos < merkle_leaves_len {
+            leaf_pos += 1;
+            merkle_leaves[leaf_pos - 1]
+        } else {
+            hash_pos += 1;
+            hashes.get(hash_pos - 1).unwrap()
+        };
+
+        let b = if proof_flags[iterator] {
+            if leaf_pos < merkle_leaves_len {
```
matt-user, line 67: You could consider adding a helper function such as get_hash(...) which returns either a leaf hash or a computed hash.  Although you would only use it twice so it may be more trouble than its worth.

matt-user (changes requested):

bitzoic:
Blocked by https://github.com/FuelLabs/sway/issues/2594. 

For now, I've added a manual test where I hash the values myself instead of using [fuel-merkle](https://github.com/FuelLabs/fuel-merkle). In this test, nodes are computed using a `u64` as opposed to a `u8`. This test case passes.

Braqzen (commented):
1. `README.md`
    - The opening sentence is duplicated in the use cases section of the spec. I don't think that this duplication is needed. I think that it makes more sense to have that sentence in the readme rather than the spec too. Furthremove, a short explanation of what the components of the merkle tree are and how these parts can be used (leafs, nodes, roots) would be beneficial.
    - I've tried to make 2 suggestions however github is getting confused with the parsing. There are 3 sections that use a single backtick ( ` ) instead of the full code block while other sections use code blocks. Could you make the following 3 sections into full code blocks?
        - `sway_libs = { git = "https://github.com/FuelLabs/sway-libs", branch = "master" }`
        - `use sway_libs::binary_merkle_proof::verify_proof`
        - `fuel-merkle = { version = "0.3" }`
    - There is a comment regarding pointing at a master branch of the repo. I suggest creating an issue to track this in advance
    - I'm unsure about the `verify_proof_example` code block being hardcoded into the readme and having just that 1 example there. It also has the postfix of `_example` at the end for some reason. I would either delete the section of change it somehow. I'm leaning towards the change rather than removal but it's not immediately obvious what I'd do here.
    - When generating a tree there is a loop and inside that loop some data is pushed into the tree while assigning the return output to a variable that is disregarded. Should that assignment be shown or can we just push without assignment?
2. `SPECIFICATION.md`
    - There's a bit of duplication in regards to the hashing. It's mentioned twice in the notes and something similar is also in the README. Perhaps a temporary section should be added where all the relevant comments are collated so that there is no duplication? I would probably put it in the README since the SPECIFICATION is the next step to read after the README if you care about the content, or even click on it in the first place.
3. Could you add [syntax highlighting?](https://github.com/FuelLabs/sway-applications/blob/master/.gitattributes)
4. `binary_merkle_proof.sw`
    - I have not mapped out the steps in `path_length_from_key` however that `while true` is worrisome. I hope that there are sufficient tests to make sure that each part of the control flow is tested to make sure that it definitely breaks out when it's meant to. I'm no expert here so I hope you can add / have added tests to make sure that this happens
    -  I think there might be something wrong with CI or the compiler since line 100 has a return expression without the use of a semi-colon at the end. Is this valid? I don't think it is. Can you verify that this builds and works as expected? Should an issue be raised?
5. Testing
    - I would split the tests into their own function modules similar to how sway-apps has done it
    - I don't really know what's going on with the tests so I've only skim read them however given my comments about the use of loops I'm unsure if all the conditions are tested (simply by looking at the line length and inferring that somewhere around this many lines of code would be just for testing the intended functionality rather than coverage testing).

bitzoic:
>    * I have not mapped out the steps in `path_length_from_key` however that `while true` is worrisome. I hope that there are sufficient tests to make sure that each part of the control flow is tested to make sure that it definitely breaks out when it's meant to. I'm no expert here so I hope you can add / have added tests to make sure that this happens

The function itself is not public and will only ever be accessed by `process_proof`. There are checks ahead of calling this function to ensure good data is being passed to it. While my testing is currently limited to do the constraint of using an array, I have a number of tests in `process_proof.rs` that use different `key` and `num_leaves` values.

>    * I think there might be something wrong with CI or the compiler since line 100 has a return expression without the use of a semi-colon at the end. Is this valid? I don't think it is. Can you verify that this builds and works as expected? Should an issue be raised?

Any return statement does not need a semi-colon in rust. The `return` keyword however is required as there is code that comes after it. No issue required.

>    * I don't really know what's going on with the tests so I've only skim read them however given my comments about the use of loops I'm unsure if all the conditions are tested (simply by looking at the line length and inferring that somewhere around this many lines of code would be just for testing the intended functionality rather than coverage testing).

As mentioned above, what I am able to test is pretty limited without the use of a Vec. Changes in the length of a proof is the only thing that is not tested here explicitly. This will be added after https://github.com/FuelLabs/fuels-rs/issues/353 is resolved.

sway_libs/src/merkle_proof/README.md
```diff
@@ -0,0 +1,92 @@
...
+### Using the Merkle Proof Library In Sway
+
+Once imported, using the Merkle Proof library is as simple as calling the desired function. Here is a list of function definitions that you may use:
+
+- `leaf_digest(data: b256) -> b256`
+- `node_digest(left: b256, right: b256) -> b256`
+- `process_proof(key: u64, merkle_leaf: b256, num_leaves: u64, proof: [b256; 2]) -> b256`
+- `verify_proof(key: u64, merkle_leaf: b256, merkle_root: b256, num_leaves: u64, proof: [b256; 2]) -> bool`
+
+> **NOTE** Fuel-RS currently does not support the Sway standard library's `vec` as a function argument. As a result, the current implementation uses an array.
```
adlerjohn, line 41:
```suggestion
> **Note** Fuels-rs does not currently support the Sway standard library's `vec` as a function argument. As a result, the current implementation uses an array. 

```

sway_libs/src/merkle_proof/README.md
```diff
@@ -0,0 +1,92 @@
...
+Once imported, using the Merkle Proof library is as simple as calling the desired function. Here is a list of function definitions that you may use:
+
+- `leaf_digest(data: b256) -> b256`
+- `node_digest(left: b256, right: b256) -> b256`
+- `process_proof(key: u64, merkle_leaf: b256, num_leaves: u64, proof: [b256; 2]) -> b256`
+- `verify_proof(key: u64, merkle_leaf: b256, merkle_root: b256, num_leaves: u64, proof: [b256; 2]) -> bool`
+
+> **NOTE** Fuel-RS currently does not support the Sway standard library's `vec` as a function argument. As a result, the current implementation uses an array. 
+
+## Using the Merkle Proof Libaray in Fuels-RS
```
adlerjohn, line 43:
```suggestion
## Using the Merkle Proof Library in Fuels-rs

```

sway_libs/src/merkle_proof/SPECIFICATION.md
```diff
@@ -0,0 +1,27 @@
...
+
+All cryptographic primitives follow the [Fuel Specs](https://github.com/FuelLabs/fuel-specs/blob/master/specs/protocol/cryptographic_primitives.md).
+
+## Hashing 
+
+All hashing is done with SHA-2-256 (also known as SHA-256), defined in [FIPS 180-4](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf).
+
+## Merkle Trees
+
+Currently only one tree structure is support, the Binary Merkle Tree. Implementation for a Binary Merkle Sum Tree and Sparse Merkle Tree found in the [fuel-merkle](https://github.com/FuelLabs/fuel-merkle) repository will be added soon.
```
simonr0204, line 21: > will be added soon

can you create an issue for this and reference it here?

sway_libs/src/merkle_proof/binary_merkle_proof.sw
```diff
@@ -0,0 +1,175 @@
...
+
+    total_length
+}
+
+/// This function will compute and return a merkle root given a leaf and corresponding proof.
+///
+/// # Arguments
+///
+/// * 'key' - The key or index of the leaf to prove.
+/// * `merkle_leaf` - The hash of a leaf on the merkle tree.
```
simonr0204, line 80: I'm going to carry on the fine tradition of insisting that "Merkle" be spelled with a capital "M" in comments. 😁

tests/src/test_projects/merkle_proof/tests/functions/leaf_digest.rs
```diff
@@ -0,0 +1,29 @@
+use crate::merkle_proof::tests::utils::{
+    abi_calls::leaf_digest,
+    test_helpers::{build_tree, merkle_proof_instance},
+};
+
+mod leaf_digest {
+
+    use super::*;
+
+    mod succes {
```
simonr0204, line 10:
```suggestion
    mod success {
```

sway_libs/src/merkle_proof/binary_merkle_proof.sw
```diff
@@ -0,0 +1,175 @@
...
+/// * When the computed height gets larger than the proof.
+pub fn process_proof(key: u64, merkle_leaf: b256, num_leaves: u64, proof: [b256;
+2]) -> b256 {
+    // let proof_length = proof.len();
+    let proof_length = 2;
+    require((num_leaves > 1 && proof_length == path_length_from_key(key, num_leaves)) || (num_leaves <= 1 && proof_length == 0), ProofError::InvalidProofLength);
+    require(key < num_leaves, ProofError::InvalidKey);
+
+    let mut digest = merkle_leaf;
+    // If there is no proof then the leaf is the root
```
simonr0204, line 98:
```suggestion
    // If proof length is zero then the leaf is the root
```

tests/src/test_projects/merkle_proof/tests/functions/process_proof.rs
```diff
@@ -0,0 +1,251 @@
+// TODO: More extensive testing should be added when https://github.com/FuelLabs/fuels-rs/issues/353 is revolved.
```
simonr0204, line 1: `fuel-merkle` has an [extensive test suite](https://github.com/FuelLabs/fuel-merkle/tree/master/tests-data
) which you might be able to use. 


The idea is for this to be a reference for all merkle tree implementations (Rust, TS, Solidity, Sway...) .

You'd need to write some wrapper code to pass the test cases into a deployed instance of your contract.

simonr0204 (changes requested):

tests/src/test_projects/merkle_proof/tests/functions/process_proof.rs
bitzoic: What I am referring to here is the need to test cases where the proof length is greater than two. This is the current array length and limits my possible tests. I've updated the comment to make that more clear.

sway_libs/src/merkle_proof/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Known Issues
+
+Due to the Fuel VM padding of copy types smaller than one word, the current hashing of the leaf and node values do not match RFC-6962. The present implementation is the equavalent of hashing a `u64` as opposed to the standard `u8`. As a result, hashing of smaller sized copy types may result in unexpected behavior. For example, the `sha-256(0u8)` will result in the same hash as `sha-256(0u64)`. This will cause an incompatability with the [Fuel-Merkle](https://github.com/FuelLabs/fuel-merkle) repository. The issue regarding this can be tracked [here](https://github.com/FuelLabs/sway/issues/2594).
+
+# Using the Library
+
+## Using the Merkle Proof Library In Sway
+
+### Importing Into Your Project
```
Braqzen, line 15: I'm unsure about this duplicated importing section. It's in the main readme and it's also in here. Perhaps the main readme should have a more generic example e.g. `use sway_libs::<library_name>::<library_function>`.
I think it's fine to have a generic explanation on the top level and then specific instructions in each library - if that is required. That being said, importing a library is the same everywhere so I'd consider removing it from here and having a generic example at the top level with perhaps 1 concrete example in case someone does not understand the generic syntax chosen by the author.

README.md
```diff
@@ -31,12 +31,19 @@ These libraries contain helper functions, generalized standards, and other tools
...
+
+You may then import your dersired library in your Sway Smart Contract as so:
+
+`use sway_libs::binary_merkle_proof::verify_proof;`
 
 ## Contributing
 
-Check [CONTRIBUTING.md](./CONTRIBUTING.md) for more info!
\ No newline at end of file
+Check [CONTRIBUTING.md](./CONTRIBUTING.md) for more info!
```
Braqzen, line 49: Note that this file is currently empty and if there isn't an issue to add to it then an issue should be created.

sway_libs/src/merkle_proof/README.md
```diff
@@ -0,0 +1,92 @@
...
+### Using the Merkle Proof Library In Sway
+
+Once imported, using the Merkle Proof library is as simple as calling the desired function. Here is a list of function definitions that you may use:
+
+- `leaf_digest(data: b256) -> b256`
+- `node_digest(left: b256, right: b256) -> b256`
+- `process_proof(key: u64, merkle_leaf: b256, num_leaves: u64, proof: [b256; 2]) -> b256`
+- `verify_proof(key: u64, merkle_leaf: b256, merkle_root: b256, num_leaves: u64, proof: [b256; 2]) -> bool`
+
+> **Note** Fuels-rs does not currently does not support the Sway standard library's `vec` as a function argument. As a result, the current implementation uses an array.
```
Braqzen, line 41: vec is under development now :)

sway_libs/src/merkle_proof/SPECIFICATION.md
```diff
@@ -0,0 +1,29 @@
...
```
Braqzen, line 21:
```suggestion
Currently only one tree structure is supported, the Binary Merkle Tree. Implementation for a Binary Merkle Sum Tree and Sparse Merkle Tree found in the [fuel-merkle](https://github.com/FuelLabs/fuel-merkle) repository will be added soon. 

```

sway_libs/src/merkle_proof/SPECIFICATION.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Hashing 
+
+All hashing is done with SHA-2-256 (also known as SHA-256), defined in [FIPS 180-4](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf).
+
+## Merkle Trees
+
+Currently only one tree structure is support, the Binary Merkle Tree. Implementation for a Binary Merkle Sum Tree and Sparse Merkle Tree found in the [fuel-merkle](https://github.com/FuelLabs/fuel-merkle) repository will be added soon. 
+
+A Sparse Merkle Tree proof can be tracked [here](https://github.com/FuelLabs/sway-libs/issues/18) and Sum Merkle Tree proof can be tracked [here](https://github.com/FuelLabs/sway-libs/issues/17);
```
Braqzen, line 23:
```suggestion
A Sparse Merkle Tree proof can be tracked [here](https://github.com/FuelLabs/sway-libs/issues/18) and Sum Merkle Tree proof can be tracked [here](https://github.com/FuelLabs/sway-libs/issues/17).

```

Braqzen (commented):
1. Could you label each codeblock so that there is syntax highlighting? This would be done by putting the word (without quotes) "rust" right after the first three backticks that open the codeblock. Note that this might have to be done without a space between the ticks and the language.
2. I would personally update the spec to list each function exposed to the user (pub) and inform them of its usage. This would be easier to follow than the ingestion of the spec imo
3. As for the testing, I would not comment out parts of a test that make it fail. I would `ignore` that test since it's clearly not finished. Commented out sections create a false impression that everything is fine until you see that the test has sections missing. Alternatively, I'd split up those test(s) if possible. Furthermore, since each function is a module I wouldn't wrap it in a `mod function_name` and instead I would use a `mod success` and `mod revert` at the top. Separating by function names was only necessary when we had many functions in 1 file.

bitzoic:
> 3. As for the testing, I would not comment out parts of a test that make it fail. I would `ignore` that test since it's clearly not finished. Commented out sections create a false impression that everything is fine until you see that the test has sections missing. 

The commented out section was left to show the use of a `u8` which causes other tests to fail, but this clearly doesn't make sense to be here. Removed and updated the comment above the test to inform the reader.

sway_libs/src/merkle_proof/binary_merkle_proof.sw
```diff
@@ -0,0 +1,175 @@
...
+    let mut num_leaves = num_leaves;
+    let mut key = key;
+
+    while true {
+        // The height of the left subtree is equal to the offset of the starting bit of the path
+        let path_length = starting_bit(num_leaves);
+        // Determine the number of leaves in the left subtree
+        let num_leaves_left_sub_tree = (1 <<(path_length - 1));
+
+        if key <= (num_leaves_left_sub_tree - 1) {
```
matt-user, line 52: Should this be a match statement?

sway_libs/src/merkle_proof/binary_merkle_proof.sw
```diff
@@ -0,0 +1,175 @@
...
+            // If the leaf is in the left subtreee, path length is full height of the left subtree
+            total_length = total_length + path_length;
+            break;
+        } else if num_leaves_left_sub_tree == 1 {
+            // If the left sub tree has only one leaf, path has one additional step
+            total_length = total_length + 1;
+            break;
+        } else if (num_leaves - num_leaves_left_sub_tree) <= 1 {
```
matt-user, line 60: Did you mean this?
```suggestion
        } else if (num_leaves - num_leaves_left_sub_tree) == 1 {
```

sway_libs/src/merkle_proof/binary_merkle_proof.sw
bitzoic: It could be, but [match guards](https://doc.rust-lang.org/1.6.0/book/patterns.html#guards) are not supported in Sway :(

sway_libs/src/merkle_proof/binary_merkle_proof.sw
bitzoic: This is intentional and is follow [fuel-merkle-sol](https://github.com/FuelLabs/fuel-merkle-sol/blob/90e8463074d9c038ef7f0b15d3bb91549812bb02/contracts/tree/Utils.sol#L41)

sway_libs/src/merkle_proof/binary_merkle_proof.sw
```diff
@@ -0,0 +1,175 @@
...
+library binary_merkle_proof;
+
+use std::{hash::sha256, revert::require};
+
+pub enum ProofError {
+    InvalidKey: (),
+    InvalidProofLength: (),
+}
+
+pub const LEAF = 0u8;
```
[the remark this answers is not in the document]
bitzoic: This is documented both in spec and comments at `leaf_digest` and `node_digest` with `MTH({d(0)}) = SHA-256(0x00 || d(0))` and `MTH(D[n]) = SHA-256(0x01 || MTH(D[0:k]) || MTH(D[k:n]))` respectively. I'll make sure to add a comment to make this more clear.

README.md
```diff
@@ -38,11 +38,21 @@ These libraries contain helper functions, generalized standards, and other tools
 To import the Merkle Proof library the following should be added to the project's `Forc.toml` file under `[dependencies]`:
 
 <!-- TODO: This should not point to the master branch but instead to a release -->
-`sway_libs = { git = "https://github.com/FuelLabs/sway-libs", branch = "master" }`
+```rust
+sway_libs = { git = "https://github.com/FuelLabs/sway-libs", branch = "master" }
+```
 
 You may then import your dersired library in your Sway Smart Contract as so:
```
Braqzen, line 45:
```suggestion
You may then import your desired library in your Sway Smart Contract as so:
```

sway_libs/src/merkle_proof/SPECIFICATION.md
```diff
@@ -8,6 +8,24 @@ It outlines the use cases, i.e. specification, and describes how to implement th
...
 
+## Public Functions
+
+### `leaf_digest`
+
+The `leaf_digest` function is used to compute leaf hash of a Merkle Tree. Given the data provided, it returns the leaf hash following [RFC-6962)](https://tools.ietf.org/html/rfc6962) as described by `MTH({d(0)}) = SHA-256(0x00 || d(0))`.
+
+### `node_digest`
+
+The `node_digest` function is used to complute a node within a Merkle Tree. Given a left and right node, it returns the node hash following [RFC-6962)](https://tools.ietf.org/html/rfc6962) as described by `MTH(D[n]) = SHA-256(0x01 || MTH(D[0:k]) || MTH(D[k:n]))`.
```
Braqzen, line 19:
```suggestion
The `node_digest` function is used to complute a node within a Merkle Tree. Given a left and right node, it returns the node hash following [RFC-6962](https://tools.ietf.org/html/rfc6962) as described by `MTH(D[n]) = SHA-256(0x01 || MTH(D[0:k]) || MTH(D[k:n]))`.

```

sway_libs/src/merkle_proof/SPECIFICATION.md
```diff
@@ -8,6 +8,24 @@ It outlines the use cases, i.e. specification, and describes how to implement th
 
 A common use case for Merkle Tree verification is airdrops. An airdrop is a method of distribution a set amount of tokens to a specified number of users. These often include a list of addresses and amounts. By posting the root hash, users can provide a proof and claim their airdrop.
 
+## Public Functions
+
+### `leaf_digest`
+
+The `leaf_digest` function is used to compute leaf hash of a Merkle Tree. Given the data provided, it returns the leaf hash following [RFC-6962)](https://tools.ietf.org/html/rfc6962) as described by `MTH({d(0)}) = SHA-256(0x00 || d(0))`.
```
Braqzen, line 15:
```suggestion
The `leaf_digest` function is used to compute leaf hash of a Merkle Tree. Given the data provided, it returns the leaf hash following [RFC-6962](https://tools.ietf.org/html/rfc6962) as described by `MTH({d(0)}) = SHA-256(0x00 || d(0))`.

```

matt-user (approved):
lgtm

Braqzen (approved):

Braqzen:
cc @simonr0204 for notification.
Github is being stupid and it's not providing an option to re-request a review from you hence the tag.

simonr0204 (approved):

PR #13, opened 2022-08-08
Add CI and tests folder

### LINKED ISSUE #3: Add CI to sway-libs
Sway-libs should have CI

bitzoic:
## Type of change

- New feature

## Changes

- Add CI
- Add build script
- Add Cargo.toml
- Add harness.rs

## Related Issues

Closes #3


Braqzen (commented):
I'm not sure that I understand what is going on here.

1. There are two test directories and there's a `.gitignore` in one rather than in the root of the repo.
2. There's some repetition between the CI jobs for the lib and tests I'm unsure if that actually installs `forc`. I was expecting something like `forc --version ${{ env.FORC_VERSION }}`. This also does not use the github action so the compilation will take a while too
3. That `run` command is a bit messy. I think it would be better to run the script with the path being passed in as an argument rather then having to `cd` twice
4. The `tests` take the `build` command but aren't run. Are these tests going to be in script form and never run? I don't understand why a `test` would be `built` but not run

bitzoic:
Most of this is based on CI from the Sway repository and following the sway-lib-std and tests.

> 1. There are two test directories and there's a `.gitignore` in one rather than in the root of the repo.

The repository structure is as follows:

- sway_libs (Not in this PR as this holds the libraries)
- tests/src
     - test_projects/
          - test_modules/ (Not in this PR as it doesn't have anything to test)
          - harness.rs
     - build.sh
     - Cargo.toml
     - .gitignore

> 2. There's some repetition between the CI jobs for the lib and tests I'm unsure if that actually installs `forc`. I was expecting something like `forc --version ${{ env.FORC_VERSION }}`. This also does not use the github action so the compilation will take a while too

build-sway-lib - Builds the libraries in sway-lib, not included in this PR. These are the libraries themselves akin to sway-lib-std.
build-test-projects - Builds the test-projects and creates the contract binaries that use the libraries from sway-lib. 
cargo-test-sway-lib - Runs the tests upon the test-projects using the binaries and .json files created from building.

I'm unsure if including `forc --version ${{ env.FORC_VERSION }}` is needed as the Sway repo follows the same structure. I can take a look at what sway-applications does.

I am not sure what you are referring to with github action.

> 3. That `run` command is a bit messy. I think it would be better to run the script with the path being passed in as an argument rather then having to `cd` twice

I am following what is done in the Sway repo but I agree, this should be fixed.

> 4. The `tests` take the `build` command but aren't run. Are these tests going to be in script form and never run? I don't understand why a `test` would be `built` but not run

Building the test projects will generate the binaries and .json needed to be run in cargo-test-sway-lib

bitzoic:
After taking a look at sway-applications I realized there was quite a lot wrong with the CI. Much of what I said in the last comment can be ignored.

I've updated it to reflect what we see in the other repo and included the sway-libs folder. A lot less repeated code now.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,67 @@
...
+          run: rustup component add rustfmt
+
+        - name: Install Fuel toolchain
+          uses: FuelLabs/action-fuel-toolchain@v0.1.0
+
+        - name: Check Sway formatting
+          run: forc fmt --path sway-libs --check
+
+        - name: Build sway-libs
+          run: forc build --path sway-libs
```
Braqzen, line 58:
```suggestion
        - name: Check Sway formatting
          run: forc fmt --path sway_libs --check

        - name: Build sway-libs
          run: forc build --path sway_libs

```

Braqzen (changes requested):
I think this might fix CI

.github/workflows/ci.yml
bitzoic: I never fail to shoot myself in the foot

PR #27, opened 2022-09-28
Bump Merkle Proof to use depth of 8

### LINKED ISSUE #26: Increase Binary Merkle Proof To Depth of 8
#### Description

The Binary Merkle Proof Library currently tests with a depth of 3. This should be increased to a depth of 8.

#### Motiviation

The current tests use an array of length 3. The standard for testing Merkle trees is a depth of 8.

bitzoic:
## Type of change

## Changes

- Use a fixed array of length 8 for binary Merkle proof
- Function to automatically return leaves for a perfect binary tree
- Function to generate a Merkle proof for a perfect binary tree

## Related Issues

Closes #26

### DIFF
--- a/sway_libs/src/merkle_proof/binary_merkle_proof.sw
+++ b/sway_libs/src/merkle_proof/binary_merkle_proof.sw
@@ -98,10 +98,10 @@ pub fn process_proof(
     key: u64,
     merkle_leaf: b256,
     num_leaves: u64,
-    proof: [b256; 2],
+    proof: [b256; 8],
 ) -> b256 {
     // let proof_length = proof.len();
-    let proof_length = 2;
+    let proof_length = 8;
     require((num_leaves > 1 && proof_length == path_length_from_key(key, num_leaves)) || (num_leaves <= 1 && proof_length == 0), ProofError::InvalidProofLength);
     require(key < num_leaves, ProofError::InvalidKey);
 
@@ -185,7 +185,7 @@ pub fn verify_proof(
     merkle_leaf: b256,
     merkle_root: b256,
     num_leaves: u64,
-    proof: [b256; 2],
+    proof: [b256; 8],
 ) -> bool {
     process_proof(key, merkle_leaf, num_leaves, proof) == merkle_root
 }
--- a/tests/src/test_projects/merkle_proof/src/main.sw
+++ b/tests/src/test_projects/merkle_proof/src/main.sw
@@ -9,14 +9,14 @@ abi MerkleProofTest {
         key: u64,
         merkle_leaf: b256,
         num_leaves: u64,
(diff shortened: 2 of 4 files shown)


simonr0204:
I may be out of the loop here, but obvious question: why 8?  That's still only 256 leaves. Are smaller trees supported (with empty values in the proof array) ?

bitzoic:
> I may be out of the loop here, but obvious question: why 8? That's still only 256 leaves. Are smaller trees supported (with empty values in the proof array) ?

It was requested by @SilentCicero to bump the depth to the standard of 8. 

It was previously only a depth of 3.

simonr0204:
> > I may be out of the loop here, but obvious question: why 8? That's still only 256 leaves. Are smaller trees supported (with empty values in the proof array) ?
> 
> It was requested by @SilentCicero to bump the depth to the standard of 8.
> 
> It was previously only a depth of 3.

Just to check - fundamentally this is just because `Vec` is not supported yet, right? The term "standard depth" has me a little confused.

bitzoic:
> > > I may be out of the loop here, but obvious question: why 8? That's still only 256 leaves. Are smaller trees supported (with empty values in the proof array) ?
> > 
> > 
> > It was requested by @SilentCicero to bump the depth to the standard of 8.
> > It was previously only a depth of 3.
> 
> Just to check - fundamentally this is just because `Vec` is not supported yet, right? The term "standard depth" has me a little confused.

Yes `Vec` is not yet supported by the SDK so a fixed length array is needed to be used. I would like to add test cases for multiple proof sizes when they are supported. It's not so much a standard but a better test as far as I'm aware.

I am also building the tree manually as the `sha256` in Sway pads `u8`as to be the same size as `u64`. I created an issue for it here: https://github.com/FuelLabs/sway/issues/2594

sway_libs/src/merkle_proof/binary_merkle_proof.sw
```diff
@@ -98,10 +98,10 @@ pub fn process_proof(
     key: u64,
     merkle_leaf: b256,
     num_leaves: u64,
-    proof: [b256; 2],
+    proof: [b256; 8],
```
matt-user, line 101: Why was this changed from 2->8 instead of 3->8?

matt-user (approved):
lgtm, just have a question regarding the change in proof depth

Braqzen (approved):
Concept ACK

sway_libs/src/merkle_proof/binary_merkle_proof.sw
bitzoic: The tree I previously had in place wasn't a perfect binary Tree and only required 2 proofs.

PR #42, opened 2022-10-27
NFT Library

### LINKED ISSUE #19: NFT Library
#### Description

In many of today's NFT project we see a standardization of a specific format for NFTs. An example of this format is the ERC-721 standard for solidity. This standardization allows for interoperability between projects and quick deployment. The Sway-Libs repository should provide such a standard for the Fuel Network.

#### Motivation

A NFT library should exist to allow developers to easily import and create their own NFTs.

#### Specification

The NFT library should follow the existing [Sway-Application NFT](https://github.com/FuelLabs/sway-applications/tree/master/NFT).

bitzoic:
## Type of change

- New feature

## Changes

- Creates a modular NFT library with the following core attributes:
    - `approve()` - Approves a user to transfer the token on the owner's behalf
    - `approved()` - Returns the approved user for the token
    - `balance_of()` - Returns the number of tokens a user owns
    - `is_approved_for_all()` - Returns whether a user is approved to transfer all tokens on another user's behalf
    - `mint()` - Instantiates a new token
    - `owner_of()` - Returns the owner of a token
    - `set_approval_for_all()` - Gives a user the ability to transfer all tokens on another user's behalf
    - `token_minted()` - Returns the number of tokens that have been minted
    - `transfer_to()` - Transfer a token from one owner to another
- The following extensions are available:
    - Administrator
         - `admin()` - Returns the administrator
         - `set_admin()` - Sets an administrator
    - Burnable
         - `burn()` - Deletes the token
    - Meta Data
         - `meta_data()` - Returns the token's stored metadata
         - `set_meta_data()` - Store new metadata for the token
    - Supply
        - `max_supply()` - Returns the maximum set supply
        - `set_max_supply()` - Assigns the maximum supply

## Notes

- More extensions may be available in the future if they are deemed common additions to NFTs
- This PR bumps to repo from forc v0.30.0 to forc v0.30.1 for the latest formatter bug fix

## Related Issues

Closes #19


bitzoic:
This is currently blocked by https://github.com/FuelLabs/sway/issues/3287

sway_libs/src/nft/SPECIFICATION.md
```diff
@@ -0,0 +1,91 @@
+# Overview
```
Braqzen, line 1: I think this doc would benefit from a ToC (table of contents)

sway_libs/src/nft/extensions/administrator/administrator_errors.sw
```diff
@@ -0,0 +1,4 @@
+library administrator_errors;
+pub enum AdminError {
```
Braqzen, line 2:
```suggestion
library administrator_errors;

pub enum AdminError {
```

sway_libs/src/nft/extensions/administrator/administrator_events.sw
```diff
@@ -0,0 +1,5 @@
+library administrator_events;
+pub struct AdminEvent {
```
Braqzen, line 2:
```suggestion
library administrator_events;

pub struct AdminEvent {
```

sway_libs/src/nft/extensions/burnable/burnable_events.sw
```diff
@@ -0,0 +1,7 @@
+library burnable_events;
+pub struct BurnEvent {
```
Braqzen, line 2:
```suggestion
library burnable_events;

pub struct BurnEvent {
```

sway_libs/src/nft/extensions/meta_data/meta_data.sw
```diff
@@ -0,0 +1,68 @@
...
+
+use meta_data_structures::NFTMetaData;
+use ::nft::{errors::InputError, nft_core::NFTCore, nft_storage::{META_DATA, TOKENS}};
+use std::{hash::sha256, storage::{get, store}};
+
+pub trait MetaData<T> {
+    /// Returns the metadata for this token
+    #[storage(read)]
+    fn meta_data(self) -> Option<T>;
+    /// Creates new metadata for this token.
```
Braqzen, line 13:
```suggestion

    /// Creates new metadata for this token.
```

sway_libs/src/nft/extensions/meta_data/meta_data.sw
```diff
@@ -0,0 +1,68 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `metadata` - The metadata value to be included in the new metadata.
+    #[storage(write)]
+    fn set_meta_data(self, metadata: Option<T>);
+}
+
+impl<T> MetaData<T> for NFTCore {
```
Braqzen, line 22: How come this is implemented on the core instead of the meta data struct?

Braqzen (commented):
I've also noticed that there is an error message (somewhere) that uses a string instead of an enum while possibly all the other errors use enums

sway_libs/src/nft/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+In order to use the `NFT` library, Sway-libs must be added to the Forc.toml file and then imported into your Sway project. To add Sway-libs as a dependency to the Forc.toml in your project please see the [README.md](../../../README.md).
+
+```rust
+use sway_libs::nft::*;
+```
+
+## Basic Functionality
+
+Once imported, a `NFT` can be minted by calling the `mint` function.
```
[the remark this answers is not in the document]
matt-user: I'm like 95% sure "a" is correct here

sway_libs/src/nft/README.md
```diff
@@ -0,0 +1,71 @@
+## Overview
+
+A non-fungible token (NFT) is a unique token that has an identifier which distinguishes itself from other tokens within the same token contract. Unlike Fuel's Native Assets, these tokens are not fungible with one another and may contain metadata or other traits giving them distinctive characteristics.
+
+While NFTs are commonly associated with artwork / collectibles, there are many greater utilities beyond that which have yet to be written for the Fuel Network.
```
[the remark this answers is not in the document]
matt-user: Totally up to you but if you wanted to you could provide an example such as [short positions](https://opyn.gitbook.io/squeeth/contracts/core-contracts)

sway_libs/src/nft/extensions/burnable/burnable.sw
```diff
@@ -0,0 +1,50 @@
...
+    fn burn(self);
+}
+
+impl Burnable for NFTCore {
+    #[storage(read, write)]
+    fn burn(self) {
+        require(self.owner == msg_sender().unwrap(), AccessError::SenderNotOwner);
+
+        let owner = self.owner;
+        let token_id = self.token_id;
```
matt-user, line 25: IMO it's a little confusing to assign variables here and then not use them immediately below where they could be used.  It also seems like they are only defined in order to make use of the field short hand syntax (or whatever it is called) ... not sure how I feel about that

sway_libs/src/nft/extensions/burnable/burnable.sw
```diff
@@ -0,0 +1,50 @@
...
+            token_id,
+        });
+    }
+}
+
+/// Burns the specified token.
+///
+/// # Arguments
+///
+/// * `token_id` - The id of the token which is to be burned.
```
matt-user, line 40:
```suggestion
/// * `token_id` - The id of the token to burn.
```
imo this is more concise

sway_libs/src/nft/extensions/meta_data/meta_data.sw
```diff
@@ -0,0 +1,68 @@
...
+
+pub trait MetaData<T> {
+    /// Returns the metadata for this token
+    #[storage(read)]
+    fn meta_data(self) -> Option<T>;
+    /// Creates new metadata for this token.
+    ///
+    /// # Arguments
+    ///
+    /// * `metadata` - The metadata value to be included in the new metadata.
```
matt-user, line 17: After looking at the implementation "included" isn't quite the right word imo.  IMO this line should read "The new metadata value to overwrite the old value." or something like that

sway_libs/src/nft/extensions/burnable/burnable.sw
```diff
@@ -0,0 +1,50 @@
+library burnable;
+
+dep burnable_events;
+
+use burnable_events::BurnEvent;
+use ::nft::{errors::{AccessError, InputError}, nft_core::NFTCore, nft_storage::{BALANCES, TOKENS}};
+use std::{auth::msg_sender, hash::sha256, logging::log, storage::{get, store}};
+
+pub trait Burnable {
+    /// Deletes this token from storage and decrements the balance of the owner.
```
matt-user, line 10: Should we also decrement from MAX_SUPPLY?

matt-user (commented):
Overall looks pretty good.  I just have some questions and some nits you can feel free to ignore

sway_libs/src/nft/extensions/burnable/burnable.sw
bitzoic: MAX_SUPPLY never changes, it is static. If there is a maximum supply of 10k, then when one burns a token the max supply is still 10k but only 9,999 exist. 

I've also implemented each extension so as to not interact with one. The burn, metadata, supply, and admin extensions are isolated.

However, where this does come into play for example is the burn functionality. It does remove the token from storage, but it does not remove the metadata. This would have to be done in a separate call implemented by the developer. 

Another example is in the mint function, there are no checks for the administrator or the supply. It will mint even if exceeding the max supply. If the check is not added by the developer using the library before calling the mint function, there is nothing preventing a new token from being minted. 

On one hand, this ensures that NFT libraries without these functionalities don't have additional gas overhead costs for checks that don't apply. On the other hand, devs now need to add these themselves. I've given them the tools but they need to use them properly.

I have been uncertain as to whether the modules should overlap or not, but have opted for the latter. To avoid the additional gas costs and ensure composability without running into any dependency problems down the road this is the approach I took. Thoughts? @matt-user @Braqzen

sway_libs/src/nft/extensions/meta_data/meta_data.sw
bitzoic: All traits should be implemented for the NFTCore and not any external structs. The example matadata in this library isn't particularly useful and each project will likely have its own metadata. 

This way, devs can just call `my_token.set_meta_data(my_struct)` with any struct they create themselves without needing to implement their own metadata get and set functions. The trait for the core takes care of this.

sway_libs/src/nft/extensions/burnable/burnable.sw
matt-user: I see that makes sense.  It's tough to say, but I think your implementation of the latter is probably the right way to go about it

matt-user (dismissed):
lgtm but I would wait for another set of eyes

sway_libs/src/nft/extensions/burnable/burnable.sw
Braqzen: I would opt for the approach of letting the developer defensively code. I'd do it on a case-by-case basis to see what makes sense. I think it's better to have a barebones implementation and allow the dev to build in the safety i.e. you can mint an unlimited about but the dev needs to ensure that you don't. It's tricky, without writing the whole thing myself and considering all of the work in unison I cannot say much more because "it depends".

sway_libs/src/nft/README.md
```diff
@@ -2,18 +2,23 @@
 
 A non-fungible token (NFT) is a unique token that has an identifier which distinguishes itself from other tokens within the same token contract. Unlike Fuel's Native Assets, these tokens are not fungible with one another and may contain metadata or other traits giving them distinctive characteristics.
 
-While NFTs are commonly associated with artwork / collectibles, there are many greater utilities beyond that which have yet to be written for the Fuel Network.
+The current widlely used case for NFTs is artwork / collectibles such as 10k minted token collections. However, there are other utilities which have yet to be written for the Fuel Network which include DeFi protocol short positions, deeds, and more.
```
Braqzen, line 5:
```suggestion
Some common applications of an NFT include artwork / collectibles, DeFi short positions, deeds, and more.
```

sway_libs/src/nft/extensions/meta_data/meta_data.sw
```diff
@@ -10,11 +10,12 @@ pub trait MetaData<T> {
     /// Returns the metadata for this token
     #[storage(read)]
     fn meta_data(self) -> Option<T>;
+
     /// Creates new metadata for this token.
     ///
     /// # Arguments
     ///
-    /// * `metadata` - The metadata value to be included in the new metadata.
+    /// * `metadata` - The new metadata to overwrite the existing metadata.
```
Braqzen, line 18: This is not useful. If I do not know what `metadata` is meant to be then telling me that it's "metadata that overwrites the old one" doesn't convey any semantic information about what `metadata` actually is.

bitzoic: Good point

PR #56, opened 2022-11-21
NFT ABI

### LINKED ISSUE #55: NFT ABI
#### Description

Every NFT should implement the same functions and thus implement the same ABI. There should be an ABI with function definitions for these base functions. 

#### Motivation

Developers should be able to import a NFT ABI rather than create their own.

bitzoic:
## Type of change

## Changes

- Add NFT ABI
- Add Supply ABI
- Add Administrator ABI
- Add Burn ABI
- Rename metadata to token_metadata to differentiate between contract and individual token metadata

## Notes

- All NFT implementations should follow the same ABI, meaning we should provide this ABI in the library.

## Related Issues

Closes #55


matt-user (commented):
In the current implementation it looks like we can set the admin to Option::None.  Is this the desired behavior?  Are there use cases for this scenario?  If not we might want to rethink this functionality

sway_libs/src/nft/nft_storage.sw
```diff
@@ -6,7 +6,7 @@ pub const ADMIN: b256 = 0x100000000000000000000000000000000000000000000000000000
 pub const APPROVED: b256 = 0x2000000000000000000000000000000000000000000000000000000000000000;
 pub const BALANCES: b256 = 0x3000000000000000000000000000000000000000000000000000000000000000;
 pub const MAX_SUPPLY: b256 = 0x4000000000000000000000000000000000000000000000000000000000000000;
-pub const META_DATA: b256 = 0x5000000000000000000000000000000000000000000000000000000000000000;
+pub const METADATA: b256 = 0x5000000000000000000000000000000000000000000000000000000000000000;
```
matt-user, line 9: I would consider renaming this TOKENMETADATA to be consistent with the rest of the code

simonr0204:
> In the current implementation it looks like we can set the admin to Option::None. Is this the desired behavior? Are there use cases for this scenario? If not we might want to rethink this functionality

I think the type should be `Option<Identity>` to handle the case where the admin hasn't been set yet, but the function to set the owner should just take an `Identity` and wrap that in an option before storing.

bitzoic:
> > In the current implementation it looks like we can set the admin to Option::None. Is this the desired behavior? Are there use cases for this scenario? If not we might want to rethink this functionality
> 
> I think the type should be `Option<Identity>` to handle the case where the admin hasn't been set yet, but the function to set the owner should just take an `Identity` and wrap that in an option before storing.

I've originally implemented this where the function takes an `Option` in order to give the ability to revoke an admin. I believe this can be a valuable feature but if it's better to only take an `Identity` then I can make the change.

> Also why do we need an abi and trait for Burn for example?

Burn is not part of the core functionality and is an extension. The trait gives the functionality to the core struct and the ABI allows to easily implement the functionality in a developer's contract. It also defines the function definition to ensure that other contracts may call it as well. Here's an example of using the ABI: https://github.com/FuelLabs/sway-libs/blob/aff22b4c998255638998e2de7e9b89789704c2ff/tests/src/test_projects/nft_extensions/src/main.sw#L56

matt-user:
To revoke admin access could the user just call set admin with the zero address or something?  Although in that case we would still interpret the NFT as having an admin and it might be a source of bugs.

sway_libs/src/nft/SPECIFICATION.md
```diff
@@ -17,7 +17,7 @@ Table of Contents
             - [`set_admin()`](#set-admin)
         - [Burnable](#burnable)
             - [`burn()`](#burn)
-        - [Metadata](#metadata)
+        - [Token Metadata](#metadata)
```
Braqzen, line 20: Some of the links in the ToC do not work given the renaming. I suggest going through the entire table to see if all of the links take you to the relevant section.

Braqzen (approved):

PR #60, opened 2022-12-14
Unused variables and private functions used - Fix

SwayStar123:
These unused variables were giving me annoying warnings on unrelated files and the private functions used made it impossible to compile the test contracts

### DIFF
--- a/tests/src/test_projects/signed_i64/src/main.sw
+++ b/tests/src/test_projects/signed_i64/src/main.sw
@@ -8,15 +8,14 @@ fn main() -> bool {
     assert(res == I64::from(2u64));
 
     res = I64::from(10u64) - I64::from(11u64);
-    assert(res == I64::from_uint(9223372036854775807u64));
-
+    assert(res == I64 { underlying: 9223372036854775807u64 });
     res = I64::from(10u64) * I64::neg_from(1);
     assert(res == I64::neg_from(10));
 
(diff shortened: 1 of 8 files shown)


tests/src/test_projects/signed_i64/src/main.sw
```diff
@@ -8,15 +8,14 @@ fn main() -> bool {
...
-
+    assert(res == I64 { underlying: 9223372036854775807u64 });
     res = I64::from(10u64) * I64::neg_from(1);
     assert(res == I64::neg_from(10));
 
     res = I64::from(10u64) * I64::from(10u64);
     assert(res == I64::from(100u64));
 
-    res = I64::from(10u64) / I64::from_uint(9223372036854775807u64);
+    res = I64::from(10u64) / I64 { underlying: 9223372036854775807u64 };
```
Braqzen, line 18: This looks like it still has a `::from()`, I assume others do as well. Why go back to regular struct syntax?

SwayStar123: Honestly I dont remember

SwayStar123: Okay so the problem is with from_uint, not from, as from_uint is a private function it shouldnt be able to be called from here, so couldnt compile the scripts on my machine due to this

Braqzen (approved):

PR #68, opened 2023-01-04
Add utility functions and tests to signed integers

chlenc:
## Type of change

- Bug fix
- New feature

## Changes

The following changes have been made to I128 and I256:

- added functions 
> fn from_u64(value: u64) -> I128 
> fn as_u64(self) -> u64 
> fn flip(self) -> Self 
> fn ge(self, other: Self) -> bool 
> fn le(self, other: Self) -> bool 
> fn zero() -> I128 
- A tests
- Added check to test not to be failed

The following changes have been made to I8, I16, I32 and I64:

- added functions 
> fn flip(self) -> Self 
> fn ge(self, other: Self) -> bool 
> fn le(self, other: Self) -> bool 
- A tests
- Added check to test not to be failed

## Notes
But my tests are still failing because of the problem I have mentioned before in this topic
https://forum.fuel.network/t/cannot-use-from-uint-in-the-i128-from-sway-libs/1255

### DIFF
--- a/sway_libs/src/signed_integers/i128.sw
+++ b/sway_libs/src/signed_integers/i128.sw
@@ -19,6 +19,12 @@ impl I128 {
             lower: 0,
         }
     }
+
+    pub fn zero() -> Self {
+        Self {
+            underlying: U128::from((0, 0)),
+        }
+    }
 }
 
 impl From<U128> for I128 {
@@ -51,6 +57,28 @@ impl core::ops::Ord for I128 {
 }
 
 impl I128 {
+    pub fn ge(self, other: Self) -> bool {
+        self > other || self == other
+    }
+
+    pub fn le(self, other: Self) -> bool {
+        self < other || self == other
+    }
+
+    pub fn from_u64(value: u64) -> I128 {
+        Self {
+            underlying: U128::from((0, value)),
+        }
+    }
+
+    pub fn as_u64(self) -> u64 {
+        if self.underlying < Self::indent() {
+            revert(0)
+        } else {
+            self.underlying.as_u64().unwrap()
+        }
+    }
+
     /// The size of this type in bits.
     pub fn bits() -> u32 {
         128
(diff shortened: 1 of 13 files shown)


sway_libs/src/signed_integers/i128.sw
```diff
@@ -169,3 +197,11 @@ impl core::ops::Subtract for I128 {
...
     }
 }
+
+impl I128 {
+     pub fn flip(self) -> Self {
+        self * Self {
+            underlying: Self::indent() - self.underlying,
+        }
+    }
+}
```
supiket, line 207:
```suggestion
}

```

tests/src/test_projects/signed_i128/tests/mod.rs
```diff
@@ -16,6 +16,8 @@ mod success {
 
         let instance = Testi128::new(wallet, path_to_bin);
 
-        let _result = instance.main().call().await;
+        let params = TxParameters::new(Some(1), Some(10000000), None);
```
supiket, line 19: Same suggestion for other instances of this, for readability purposes.

```suggestion
        let params = TxParameters::new(Some(1), Some(10_000_000), None);

```

tests/src/test_projects/signed_i128/tests/mod.rs
```diff
@@ -16,6 +16,8 @@ mod success {
 
         let instance = Testi128::new(wallet, path_to_bin);
 
-        let _result = instance.main().call().await;
+        let params = TxParameters::new(Some(1), Some(10000000), None);
+        let _result = instance.main().tx_params(params).call().await;
+        assert_eq!(_result.is_err(), false);
```
supiket, line 21: The underscore is used for unused variables.
Same suggestion for other instances of this usage.

```suggestion
        let result = instance.main().tx_params(params).call().await.unwrap();
        assert!(result.value);

```

supiket (changes requested):
These are my comments for this stage of the work, but since this PR is a work in progress, I will convert this _open_ PR to _draft_. I also edited the title to reflect that these changes are implemented for all signed integer types. You can mark it _ready for review_ (_open_) once functions for other types and tests for all the new functions are added.

PR #76, opened 2023-01-12
Contract Owner Library

### LINKED ISSUE #53: Contract Owner Library
#### Description

In Solidity, a common library that is used is the `ownerOf` modifier. This allows for developers to ensure a function cannot be called unless a specific user is the caller.

Sway should have a similar feature that easily allows for developers to setup an owner of a contract.

#### Motivation

Contracts often require owner privileges.

bitzoic:
## Type of change

- New feature

## Changes

- Added a small library to import and use ownership in a contract, the following functions have been added:
    - `only_owner()`
    - `owner()`
    - `renounce_ownership()`
    - `set_ownership()`
    - `transfer_ownership()`
- Added README.md for this library
- Added SPECIFICATION.md for this library 
- Added tests for this library

## Notes

- It may be useful to update some of the apps in the sway-applications repo to use this library, using the `as` keyword to change "owner" to "admin" as necessary

## Related Issues

Closes #53


sway_libs/src/ownership/ownable.sw
```diff
@@ -0,0 +1,52 @@
...
+    require(get::<Option<Identity>>(OWNER).is_none(), AccessError::OwnerExists);
+    store(OWNER, Option::Some(new_owner));
+
+    log(OwnershipSet { new_owner });
+}
+
+#[storage(read, write)]
+pub fn transfer_ownership(new_owner: Identity) {
+    let previous_owner = get::<Option<Identity>>(OWNER);
+    only_owner();
```
matt-user, line 45:
```suggestion
    only_owner();
    let previous_owner = get::<Option<Identity>>(OWNER);
```
saves a storage read in the case the caller is not the owner

sway_libs/src/ownership/ownable.sw
```diff
@@ -0,0 +1,52 @@
...
+
+#[storage(read)]
+pub fn owner() -> Option<Identity> {
+    get::<Option<Identity>>(OWNER)
+}
+
+#[storage(read, write)]
+pub fn renounce_ownership() {
+    let previous_owner = get::<Option<Identity>>(OWNER);
+    only_owner();
```
matt-user, line 26:
```suggestion
    only_owner();
    let previous_owner = get::<Option<Identity>>(OWNER);
```
saves a storage read in the case the caller is not the owner

sway_libs/src/ownership/ownable.sw
```diff
@@ -0,0 +1,52 @@
...
+    require(owner.is_some() && msg_sender().unwrap() == owner.unwrap(), AccessError::NotOwner);
+}
+
+#[storage(read)]
+pub fn owner() -> Option<Identity> {
+    get::<Option<Identity>>(OWNER)
+}
+
+#[storage(read, write)]
+pub fn renounce_ownership() {
```
matt-user, line 24: I don't see the purpose of this function.  If an owner renounces ownership of a contract then anyone can call the `set_ownership` function and gain restricted access to the contract in question.  IMO `renounce_ownership` should either set the owner to the 0 address and render all restricted access to the contract useless (if we do this we should add warnings to the code and docs).  Or `renounce_ownership` should be removed and restricted access can only be changed by using the `transfer_ownership` function

matt-user (changes requested):
I have an issue with the `renounce_ownership` function but other than that it lgtm

sway_libs/src/ownership/ownable.sw
bitzoic: Good point. Instead of reading from memory again and creating a new variable, I now call `msg_sender()` since we've already verified that the sender is the owner.

sway_libs/src/ownership/ownable.sw
bitzoic: This is definitely something I thought of when writing this. I would like to keep `renounce_ownership` as it follows openzeppilin's implementation of ownership and is something that Solidity developers are familiar with. 

Since this function should be called in the developer's constructor, in practice it shouldn't ever be called again... But it does leave the door open for them to shoot themselves in the foot.

I would also like to avoid the use of the 0 address as calling `owner()` should return `None`. Adding an enum to ensure that the `set_ownership()` function cannot be called twice is another approach but will incur more gas costs.

Would like to hear input from others as well on this

sway_libs/src/ownership/ownable.sw
```diff
@@ -0,0 +1,50 @@
...
+
+use errors::AccessError;
+use events::{OwnershipRenounced, OwnershipSet, OwnershipTransferred};
+use ownable_storage::OWNER;
+use std::{auth::msg_sender, hash::sha256, logging::log, storage::{get, store}};
+
+#[storage(read)]
+pub fn only_owner() {
+    let owner = get::<Option<Identity>>(OWNER);
+    require(owner.is_some() && msg_sender().unwrap() == owner.unwrap(), AccessError::NotOwner);
```
simonr0204, line 15: I think the 2nd condition here is sufficient.

simonr0204:
Does this work? I didn't think libraries could read/write contract storage.

sway_libs/src/ownership/ownable.sw
simonr0204: Non-initialized ownership and burned ownership are two distinct states - anyone using this library should be forced to handle them explicitly and separately. 

Maybe ownership could be an `enum` instead of an `option`, containing 3 variants for uninitialized, set, and burned ownership. We could provide some helper functions to `unwrap` the various cases.

matt-user: I agree with @simonr0204 that we should have 3 variants for uninitialized, set, and renounced(burned) ownership.  Looking at the openzeppelin docs their implementation of `renounce_ownership` makes it so no one can set ownership after the function is called.  We should also have this behavior.

bitzoic: I've updated the library accordingly

sway_libs/src/ownership/ownable.sw
bitzoic: Correct me if I'm wrong but I believe this is needed. If the owner is `None` (In the cases of initialized or renounced) then we cannot unwrap it

matt-user (dismissed):
lgtm but you may want to wait for another review

simonr0204:
I was wondering whether it would be cleaner to combine the notions of `State` and `Owner` into something like:

```
pub enum Owner {
     Uninitialized: (),
     Initialized: Identity,
     Revoked: (),
 }
```

Then instead of `unwrap` on the owner, you would have a match statement on this enum.

bitzoic:
> I was wondering whether it would be cleaner to combine the notions of `State` and `Owner` into something like:
> 
> ```
> pub enum Owner {
>      Uninitialized: (),
>      Initialized: Identity,
>      Revoked: (),
>  }
> ```
> 
> Then instead of `unwrap` on the owner, you would have to match on this enum, and handle each of the possible owner states. IMO this is fine (and desirable) since any app using this should be prepared to deal with the three possible states.
> 
> If we really wanted to return this as an option (i.e, force it into one of two states : Some or None) we could implement something like :
> 
> ```
> pub fn get_owner_as_option() -> Option<Identity> {
>      match owner() {
>          Owner::Initialized(identity) => Option::Some(identity),
>          _ => Option::None(),
>      }   
> )
> ```

This is wayyyy cleaner. I've updated it to use the identity in the enum

simonr0204 (approved):

matt-user (approved):

simonr0204 (approved):

PR #78, opened 2023-01-16
Refactor libraries into their own pkgs under a `libs` workspace

mitchmindtree:
I came across this while experimenting with a prototype for https://github.com/FuelLabs/sway/issues/3752 which involved depending on the `nft` lib, so I thought I'd put together a PR!

Currently, all libraries under sway-libs are combined under a single forc package. This means that when using sway-libs as a dependency, forc requires type-checking all inner libraries, even if the user only wishes to use one of the libraries.

This PR aims to split up the sway-libs package into a unique package for each inner library. A libs workspace is introduced to make building the libraries together a little easier.

**Breaking**

Note that this will break downstream packages that currently depend on sway-libs as a single library. The README has been updated to demonstrate the new approach for importing libraries from the sway-libs repo.


bitzoic (commented):
VERY excited for package hosting 🚀 

There are a few broken links now in the README of the individual libraries due to the reorganization of files, besides that LGTM!!

I'll make sure to mention the breaking changes in the next release. I made some breaking changes recently as well so I'll hold off on the release until we get this merged

mitchmindtree:
Thanks @bitzoic!

I've rebased onto master, updated the new `ownership` lib for the new workspace changes, and I think I've fixed those links and out-of-date instructions in the individual library READMEs.

bitzoic (approved):
:shipit:

PR #88, opened 2023-01-27
Update String library to use `Bytes` type

### LINKED ISSUE #82: Update strings library to use `Bytes` type
### Motivation

This would at least get rid of all the extra padding.

bitzoic:
## Type of change

## Changes

- Updated the String library to uses the `Bytes` type
- Added new functionality to the `String` type:
     - `from()`
     - `join()`
     - `set()`
     - `split()`
     - `swap()`
     - `as_vec()`
     - `into()`

## Notes

- The `as_bytes()` function no longer exists and has been replaced with the `From` trait. This is a breaking change.

## Related Issues

Closes #82


libs/string/src/lib.sw
```diff
@@ -20,13 +27,24 @@ impl String {
...
     }
 
+    /// Converts bytes to a `String`.
+    ///
+    /// # Arguments
+    ///
+    /// * `bytes` - The bytes which will be converted into a `String`.
+    pub fn from_bytes(bytes: Bytes) -> String {
+        String { bytes }
+    }
```
Braqzen, line 37:
```suggestion
    pub fn from_bytes(bytes: Bytes) -> Self {
        Self { bytes }
    }
```

libs/string/src/lib.sw
```diff
@@ -20,13 +27,24 @@ impl String {
...
     ///
     /// # Arguments
     ///
     /// * `bytes` - The vector of `u8` bytes which will be converted into a `String`.
-    pub fn from_utf8(bytes: Vec<u8>) -> String {
-        String { bytes }
+    pub fn from_utf8(mut bytes: Vec<u8>) -> String {
+        String {
+            bytes: Bytes::from_vec_u8(bytes),
+        }
```
Braqzen, line 47:
```suggestion
    pub fn from_utf8(mut bytes: Vec<u8>) -> Self {
        Self {
            bytes: Bytes::from_vec_u8(bytes),
        }
```

Does this need to be `mut`?

libs/string/src/lib.sw
```diff
@@ -89,14 +107,55 @@ impl String {
...
+    }
+
+    /// Splits a `String` at the given index, modifying the original and returning the right-hand side `String`.
+    ///
+    /// # Arguments
+    ///
+    /// * `index` - The index to split the original String at.
+    pub fn split(ref mut self, index: u64) -> String {
+        String::from_bytes(self.bytes.split(index))
+    }
```
Braqzen, line 160:
```suggestion
    pub fn join(ref mut self, other: self) -> Self {
        Self::from_bytes(self.bytes.join(other.as_bytes()))
    }

    /// Splits a `String` at the given index, modifying the original and returning the right-hand side `String`.
    ///
    /// # Arguments
    ///
    /// * `index` - The index to split the original String at.
    pub fn split(ref mut self, index: u64) -> String {
        Self::from_bytes(self.bytes.split(index))
    }
```

Does this work or is this bugged too?

Braqzen (commented):
I would add pre-conditional tests i.e. a before and after to prove that things have changed e.g. in the `set()` test.
Is there a point in the rust tests since they only call a wrapper i.e. check if something does not panic but not the actual behaviour?

libs/string/src/lib.sw
bitzoic: Yes, without it I get a `Cannot call method "from_vec_u8" on variable "bytes" because "bytes" is not declared as mutable.` as it's defined as mutable in the Bytes library. 

cc @nfurfaro

bitzoic:
> Is there a point in the rust tests since they only call a wrapper i.e. check if something does not panic but not the actual behaviour?

I would eventually like to return the `String` once that is supported by the SDK. This also follows the same format of tests for the entire repository.

libs/string/src/lib.sw
```diff
@@ -25,8 +27,10 @@ impl String {
     /// # Arguments
     ///
     /// * `bytes` - The vector of `u8` bytes which will be converted into a `String`.
-    pub fn from_utf8(bytes: Vec<u8>) -> String {
-        String { bytes }
+    pub fn from_utf8(mut bytes: Vec<u8>) -> String {
```
Braqzen, line 30:
```suggestion
    pub fn from_utf8(mut bytes: Vec<u8>) -> Self {
```

Can you confirm or deny if this works? I think it should work and I have suggested it before.

libs/string/src/lib.sw
```diff
@@ -89,14 +93,67 @@ impl String {
...
     pub fn with_capacity(capacity: u64) -> Self {
         Self {
-            bytes: Vec::with_capacity(capacity),
+            bytes: Bytes::with_capacity(capacity),
         }
     }
 }
+
+impl From<Bytes> for String {
+    fn from(b: Bytes) -> String {
```
Braqzen, line 129:
```suggestion
    fn from(b: Bytes) -> Self {
```

Does this work?

libs/string/src/lib.sw
```diff
@@ -89,14 +93,67 @@ impl String {
...
+    pub fn join(ref mut self, other: self) -> Self {
+        Self::from(self.bytes.join(other.into()))
+    }
+
+    /// Splits a `String` at the given index, modifying the original and returning the right-hand side `String`.
+    ///
+    /// # Arguments
+    ///
+    /// * `index` - The index to split the original String at.
+    pub fn split(ref mut self, index: u64) -> String {
```
Braqzen, line 156:
```suggestion
    pub fn split(ref mut self, index: u64) -> Self {
```

Does this work?

libs/string/src/lib.sw
bitzoic: Whoops, missed these. Updated 👍

Braqzen (approved):

PR #92, opened 2023-02-09
Introduce StorageString to sway-libs

### LINKED ISSUE #40: StorageString Library
#### Description

The current [`String`](https://github.com/FuelLabs/sway-libs/tree/master/sway_libs/src/string) interface is only supported in the heap. This does not allow for any persistent storage within the contract itself. There should be a `StorageString` based on the `StorageVec` in order to allow users to save their string.

#### Motivation

Some use cases require strings to be stored.

bitzoic:
## Type of change

- New feature

## Changes

- Added `StorageString` type
- Added `from_raw_slice()` and `as_raw_slice()` to the `String` type
- Updated src README wording, links, and added `StorageString`
- Moved `String` into `sway-libs/libs/strings` folder with the `StorageString` type
- Updated `String` README and SPECIFICATION to remove outdated info
- Updated `String` to use non-mutable types in `from_utf8()`

## Notes

- The `From<raw_slice>` implementation for the `String` type is commented out until https://github.com/FuelLabs/sway/issues/3637 is resolved. The `from_raw_slice()` shall be removed when this is reimplemented
- Tests for the `from_raw_slice()` and `as_raw_slice()` functions have been commented out until https://github.com/FuelLabs/sway/issues/4408 is resolved
- Until https://github.com/FuelLabs/fuels-rs/issues/940 is resolved, developers must return any  `StorageString`s from storage as a  `Bytes` type
- This should unblock NFT URIs, Fuel Name Service, token standards, and more

## Related Issues

Closes #40


libs/strings/storage_string/src/lib.sw
```diff
@@ -0,0 +1,143 @@
...
+            Option::None => Option::None,
+        }
+    }
+
+    /// Clears a stored string in storage.
+    ///
+    /// ### Number of Storage Accesses
+    ///
+    /// * Reads: `1`
+    /// * Clears: `2`
```
Braqzen, line 86: There's a read, write and a clear? What is a clear? Is it a form of a write? I would use write.

libs/strings/storage_string/src/lib.sw
```diff
@@ -0,0 +1,143 @@
...
+    ///     assert(retrieved_string.is_none());
+    /// }
+    /// ```
+    #[storage(read, write)]
+    fn clear(self) -> bool {
+        let key = __get_storage_key();
+        clear_slice(key)
+    }
+
+    /// Returns the length of string in storage.
```
Braqzen, line 115:
```suggestion
    /// Returns the length of `String` in storage.
```

libs/strings/string/src/lib.sw
```diff
@@ -137,6 +145,30 @@ impl From<Bytes> for String {
...
 
+impl AsRawSlice for String {
+    /// Returns a raw slice to all of the elements in the string.
+    fn as_raw_slice(self) -> raw_slice {
+        asm(ptr: (self.bytes.buf.ptr(), self.bytes.len)) { ptr: raw_slice }
+    }
+}
+
+
+
```
Braqzen, line 156:
```suggestion
```

tests/Forc.toml
```diff
@@ -41,6 +41,7 @@ members = [
   "./src/signed_integers/signed_i16_twos_complement",
   "./src/signed_integers/signed_i32_twos_complement",
   "./src/signed_integers/signed_i64_twos_complement",
+  "./src/storage_string",
```
Braqzen, line 44: libs/Forc.toml has this below the storagemapvec. We should be consistent and pick above / below, similarly in the harness.rs below

tests/src/storage_string/src/main.sw
```diff
@@ -0,0 +1,47 @@
...
+}
+
+impl MyContract for Contract {
+    #[storage(read, write)]
+    fn clear_string() -> bool {
+        storage.stored_string.clear()
+    }
+
+    #[storage(read)]
+    fn get_string() -> Bytes {
```
Braqzen, line 29: Wondering if this should return a Result instead

tests/src/storage_string/tests/functions/clear_string.rs
```diff
@@ -0,0 +1,26 @@
...
+    let instance = setup().await;
+
+    let string = "Fuel is blazingly fast!";
+    let input = String {
+        bytes: Bytes(string.as_bytes().to_vec()),
+    };
+
+    assert!(clear_string(&instance).await);
+
+    store_string(input.clone(), &instance).await;
```
Braqzen, line 19:
```suggestion
    store_string(input, &instance).await;
```

tests/src/storage_string/tests/functions/get_string.rs
```diff
@@ -0,0 +1,22 @@
...
+use fuels::prelude::Bytes;
+
+#[tokio::test]
+async fn gets_string_length() {
+    let instance = setup().await;
+
+    let string = "Fuel is blazingly fast!";
+    let input = String {
+        bytes: Bytes(string.as_bytes().to_vec()),
+    };
```
Braqzen, line 15: This is constructed a number of times, perhaps factor into a function?

tests/src/storage_string/tests/functions/get_string.rs
```diff
@@ -0,0 +1,22 @@
...
+
+    assert_eq!(get_string(&instance).await, Bytes(vec![]));
```
Braqzen, line 17: Wondering if the macro can be changed to a slice?

tests/src/storage_string/tests/functions/store_string.rs
```diff
@@ -0,0 +1,68 @@
...
+    Lorem ipsum dolor sit amet, consectetuer adipiscing elit. Morbi gravida libero nec velit. Morbi scelerisque luctus velit. Etiam dui sem, fermentum vitae, sagittis id, malesuada in, quam. Proin mattis lacinia justo. Vestibulum facilisis auctor urna. Aliquam in lorem sit amet leo accumsan lacinia. Integer rutrum, orci vestibulum ullamcorper ultricies, lacus quam ultricies odio, vitae placerat pede sem sit amet enim. Phasellus et lorem id felis nonummy placerat. Fusce dui leo, imperdiet in, aliquam sit amet, feugiat eu, orci. Aenean vel massa quis mauris vehicula lacinia. Quisque tincidunt scelerisque libero. Maecenas libero. Etiam dictum tincidunt diam. Donec ipsum massa, ullamcorper in, auctor et, scelerisque sed, est. Suspendisse nisl. Sed convallis magna eu sem. Cras pede libero, dapibus nec, pretium";
+    let input = String {
+        bytes: Bytes(string.as_bytes().to_vec()),
+    };
+
+    assert_eq!(stored_len(&instance).await, 0);
```
Braqzen, line 38: Some tests check for an empty bytes as a precon, this one does not. For consistency we should add it here

tests/src/storage_string/tests/functions/store_string.rs
```diff
@@ -0,0 +1,68 @@
...
+    let string1 = "Fuel is the fastest modular execution layer";
+    let string2 = "Fuel is blazingly fast!";
+    let input1 = String {
+        bytes: Bytes(string1.as_bytes().to_vec()),
+    };
+    let input2 = String {
+        bytes: Bytes(string2.as_bytes().to_vec()),
+    };
+
+    store_string(input1.clone(), &instance).await;
```
Braqzen, line 59: I would add precons here too

tests/src/storage_string/tests/functions/store_string.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    assert_eq!(get_string(&instance).await, input1.bytes);
+    assert_eq!(stored_len(&instance).await, string1.as_bytes().len() as u64);
+
+    store_string(input2.clone(), &instance).await;
+
+    assert_eq!(get_string(&instance).await, input2.bytes);
+    assert_eq!(stored_len(&instance).await, string2.as_bytes().len() as u64);
```
Braqzen, line 67: I'd add 1 more assertion to prove that the strings are different and storing one after the other results in the previous one being overwritten entirely

tests/src/storage_string/tests/functions/stored_len.rs
```diff
@@ -0,0 +1,22 @@
+use crate::storage_string::tests::utils::{
+    abi_calls::{store_string, stored_len},
+    test_helpers::setup,
+    String,
+};
+use fuels::prelude::Bytes;
+
+#[tokio::test]
+async fn clears_string() {
```
Braqzen, line 9: This name appears to be incorrect

tests/src/string/src/main.sw
```diff
@@ -268,6 +291,30 @@ impl StringTest for Contract {
         // assert(bytes.get(0).unwrap() == string.nth(0).unwrap());
     }
 
+    fn test_into_raw_slice() {
+        let mut string = String::new();
+
+        let raw_slice: raw_slice = string.as_raw_slice();
+        assert(raw_slice.number_of_bytes() == string.len());
+
+
```
Braqzen, line 300:
```suggestion
```

My disappointment is immeasurable and my day is ruined

tests/src/string/tests/functions/from_raw_slice.rs
```diff
@@ -0,0 +1,13 @@
...
+
+mod success {
+
+    use super::*;
+
+    #[tokio::test]
+    async fn converts_from_raw_slice() {
+        let instance = setup().await;
+
+        from_raw_slice(&instance).await;
```
Braqzen, line 11: I'd add some assertions here if possible

tests/src/string/tests/functions/into_raw_slice.rs
```diff
@@ -0,0 +1,13 @@
...
+
+mod success {
+
+    use super::*;
+
+    #[tokio::test]
+    async fn returns_raw_slice() {
+        let instance = setup().await;
+
+        into_raw_slice(&instance).await;
```
Braqzen, line 11: I'd add some assertions here if possible

Braqzen (commented):
There are a number of functions that return an empty `()` fuel response, which is left unused. I would consider removing the return type entirely since no logs are emitted and there is no return value or it's unused in the tests

tests/Forc.toml
bitzoic: This follows alphabetical order as does libs/Forc.toml.

tests/src/storage_string/tests/functions/store_string.rs
bitzoic: I'm not sure if I understand. You want me to add
```rust
assert!(string1 != string2);
```
when they are hardcoded?

tests/src/string/tests/functions/from_raw_slice.rs
bitzoic: Assertions are done in the contract as this is a heap type. This will be updated soon https://github.com/FuelLabs/sway-libs/issues/137

tests/Forc.toml
Braqzen: ah, I've blanked on the 1st portion of the path

tests/src/storage_string/tests/functions/store_string.rs
Braqzen: For correctness / completeness of the test we should prove that after the second storage operation the string is different. This test does not enforce this and thus I could make the strings take the same value and everything would pass.

If the test is refactored in any way we need to make sure that something does not cause a false positive and thus we must prove that the strings are different.

Tests should be written as close to the purpose of a function as reasonable i.e. I should be able to pass multiple strings into this and they should all behave as expected e.g. no false positives.

libs/strings/storage_string/README.md
```diff
@@ -0,0 +1,58 @@
...
+```
+
+## Basic Functionality
+
+Storing a `String` type can be done by calling the `store` function.
+
+```rust
+// Create a new string
+let mut my_string = String::new();
+my_string.push(0u8);
```
nfurfaro, line 45: For this basic example, could this push something other than an integer?
i.e: can "Hello" be pushed, or do we need to manually translate a string to bytes encoding to use this?

bitzoic: We would manually need to translate this to bytes unfortunately. This is where we are lacking in terms of `str` and `String` conversion.

nfurfaro: No worries; I review a little further and saw where you make this conversion in the SDK tests.

libs/strings/storage_string/README.md
```diff
@@ -0,0 +1,58 @@
...
+        <source media="(prefers-color-scheme: dark)" srcset=".docs/storage-string-logo-dark-theme.png">
+        <img alt="SwayApps logo" width="400px" src=".docs/storage-string-logo-light-theme.png">
+    </picture>
+</p>
+
+# Overview
+
+The StorageString library provides an interface to store UTF-8 encoded strings of dynamic length in Sway. The `StorageString` can be used in combination with the `String` type.
+
+The `StorageString` is stored as a collection of tightly packed bytes. This differs from Sway's built in `str` because the size cannot be known at compile time and the length is dynamic.
```
nfurfaro, line 12: Could this maybe use some distinction between this type and the StorageBytes type?
As they both store a collection of tightly packed bytes it may be confusing to users. 
I think the addition of a `char` type will help here. Using the `StorageString` now sets you up for when we have better string support in the core language.

nfurfaro (commented):
Looks good! Just a couple small comments.

libs/strings/storage_string/README.md
bitzoic: Changed to state it stores the underlying data of the `String` type

Braqzen (approved):
Personally I would increase the number of assertions to clamp down further on the tests but this is fine

PR #104, opened 2023-02-23
Optimize gas usage of node and leaf hashes

### LINKED ISSUE #95: Introduce Bitmask Merkle Proof
#### Description

The current implementation of the binary merkle proof library follows the [RFC-6962](https://www.rfc-editor.org/rfc/rfc6962) specification. This makes it compatible with the Fuel-Merkle repository and follows proper binary merkle proof spec.

However, the current approach can be gas intensive. The current implementation requries that we calculate which side of the tree the specifed leaf or node is on.

To circumvent this, it is proposed that we provide a `Vec` of bits or use a bitmask and eliminate the need for this computation. This keeps all computation off chain and still follows the spec.

#### Motivation

It was noted by internal contributors that the gas usage of the binary merkle proof library can limit the size of the merkle tree depth.

bitzoic:
## Type of change

## Changes

- Refactor leaf and node hashing to improve gas usage
- Added test for a complete but not full tree

## Notes

- This **DOES NOT** implement a bitmask described in #95. After some thorough testing, I don't think this is necessary. Current optimizations in conjunction with changes within the compiler and the stdlib have already brought gas down SIGNIFICANTLY. Million leaf merkle trees are no longer expensive. If this is something we would like to pursue in the future we can reopen the issue. 
- Some testing using the Rust SDK shows that a bitmask only provides a 0.2% decrease in gas costs.
- The recent bump to forc v0.35.3 decrease gas usage from 5,802,095 -> 26,076 for 1 million leaf merkle tree
- Current changes reduce gas usage by another 84% from master from 26,076 -> 4,017 for 1 million leaf merkle tree

## Related Issues

Closes #95

### DIFF
--- a/libs/merkle_proof/src/binary_merkle_proof.sw
+++ b/libs/merkle_proof/src/binary_merkle_proof.sw
@@ -20,11 +20,12 @@ pub const NODE = 1u8;
 ///
 /// * 'data' - The hash of the leaf data.
 pub fn leaf_digest(data: b256) -> b256 {
-    let mut bytes = Bytes::new();
-    let mut b256_as_bytes = Bytes::from(data);
+    let mut bytes = Bytes::with_capacity(33);
+    let new_ptr = bytes.buf.ptr().add_uint_offset(1);
 
-    bytes.push(LEAF);
-    bytes.append(b256_as_bytes);
+    bytes.buf.ptr().write_byte(LEAF);
+    __addr_of(data).copy_bytes_to(new_ptr, 32);
+    bytes.len = 33;
 
     bytes.sha256()
 }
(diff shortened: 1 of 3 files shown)


Braqzen (approved):
I'm unsure about manipulating things at such a low level here but for now it's an improvement.

tests/src/merkle_proof/tests/functions/process_proof.rs
```diff
@@ -29,7 +27,7 @@ mod success {
     async fn processes_merkle_proof() {
         let instance = merkle_proof_instance().await;
 
-        let depth = 8;
+        let depth = 16;
```
matt-user, line 30: should we increase depth on any of the other tests?  Doesn't have to be in this pr but something to consider.

matt-user (approved):
🔥

tests/src/merkle_proof/tests/functions/process_proof.rs
bitzoic: I've tested it with depth of 24, which takes about 10 minutes to create the tree on my local machine. This would take way too long for CI

libs/merkle_proof/src/binary_merkle_proof.sw
```diff
@@ -20,11 +20,12 @@ pub const NODE = 1u8;
 ///
 /// * 'data' - The hash of the leaf data.
 pub fn leaf_digest(data: b256) -> b256 {
-    let mut bytes = Bytes::new();
-    let mut b256_as_bytes = Bytes::from(data);
+    let mut bytes = Bytes::with_capacity(33);
```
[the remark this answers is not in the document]
bitzoic: The spec is '0x00 + b256' for leaves and '0x01 + b256 + b256' for nodes. 33 and 65 bytes respectively

PR #110, opened 2023-03-10
Two's complement for `I128` and `I256`

rostyslavtyshko:
## Type of change

- New feature

## Changes

Two's complement functions for `I128` and `I256`

Fixed warnings

## Notes

- Note 1

## Related Issues

Closes #\<issue number\>

### DIFF
--- a/libs/signed_integers/src/i16.sw
+++ b/libs/signed_integers/src/i16.sw
@@ -195,7 +195,7 @@ impl TwosComplement for I16 {
         if self.underlying >= Self::indent() {
             return self;
         }
-        let res = Self::from_uint(!self.underlying + 1);
+        let res = Self::from_uint(!self.underlying + 1u16);
         res
     }
 }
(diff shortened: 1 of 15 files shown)


bitzoic (approved):
Thanks for fixing the warnings 🎉

libs/signed_integers/src/i16.sw
```diff
@@ -195,7 +195,7 @@ impl TwosComplement for I16 {
         if self.underlying >= Self::indent() {
             return self;
         }
-        let res = Self::from_uint(!self.underlying + 1);
+        let res = Self::from_uint(!self.underlying + 1u16);
```
nfurfaro, line 198: good catch. Is there a test for this now?
Actually, it probably doesn't affect the outcome at all.

libs/signed_integers/src/i256.sw
```diff
@@ -208,18 +208,20 @@ impl core::ops::Subtract for I256 {
...
+        {
+            return self;
+        }
+        let u_one = U256 {
+            a: 0,
+            b: 0,
+            c: 0,
+            d: 1,
+        };
+        let res = I256::from_uint(!self.underlying + u_one);
```
nfurfaro, line 224: in general, you can just return `I256::from_uint(!self.underlying + u_one)` implicitly and skip creating a new temporary local variable. 
Same thing for `I128` and the rest of the types.

PR #115, opened 2023-03-16
Add fungible token abi

nfurfaro:
## Type of change

- New feature

## Changes

- Added the base abi for a fungible token on Fuel

## Notes

- authors: this manifest field could have additional entries for significant contributors?
- name: SRC20 was also suggested. I went with FRC20 initially as it seems to be common to use the network name rather than the Language name  (i.e: ERC20 uses the "E" from Ethereum, cosmos has the ICS-20 standard, etc.)

### DIFF
--- a/libs/frc20_abi/Forc.toml
+++ b/libs/frc20_abi/Forc.toml
@@ -1,0 +1,5 @@
+[project]
+authors = ["Fuel Labs <contact@fuel.sh>"]
+entry = "lib.sw"
+license = "Apache-2.0"
+name = "frc20_abi"
--- a/libs/frc20_abi/src/lib.sw
+++ b/libs/frc20_abi/src/lib.sw
@@ -3,10 +3,8 @@ library frc20_abi;
 use std::u256::U256;
 
 abi FRC20 {
-    #[storage(read)]
     /// Get the total supply of the token.
-    /// MUST track all mint and burn operations.
-    /// /// /// MAY store the amount as another type (i.e: 'u64') internally.
+    #[storage(read)]
     fn total_supply() -> U256;
 
     /// Get the name of the token
--- a/libs/frc20_abi/src/lib.sw
+++ b/libs/frc20_abi/src/lib.sw
@@ -1,0 +1,22 @@
+library frc20_abi;
+
+use std::u256::U256;
+
+abi Token {
+    #[storage(read)]
+    /// Get the total supply of the token.
+    /// MUST track all mint and burn operations.
+    /// MAY store the amount as another type (i.e: 'u64') internally.
+    fn total_supply() -> U256;
+
+    /// Get the name of the token
+    /// Example (with trailing padding): "MY_TOKEN                                                        "
+    fn name() -> str[64];
+
+    /// Get the symbol of the token
+    /// Example (with trailing padding): "TKN                             "
+    fn symbol() -> str[32];
+
+    /// Get the decimals of the token
+    fn decimals() -> u8;
+}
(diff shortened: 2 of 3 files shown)


libs/frc20_abi/src/lib.sw
```diff
@@ -0,0 +1,22 @@
+library frc20_abi;
+
+use std::u256::U256;
+
+abi Token {
+    #[storage(read)]
+    /// Get the total supply of the token.
+    /// MUST track all mint and burn operations.
+    /// MAY store the amount as another type (i.e: 'u64') internally.
```
Braqzen, line 9:
```suggestion
    /// Get the total supply of the token.
    /// MUST track all mint and burn operations.
    /// MAY store the amount as another type (i.e: 'u64') internally.
    #[storage(read)]
```

Unsure about the last 2 comments.

What is the purpose of telling the reader that it must track both? Under which conditions would that not be the case? If anything this should only be a getter, and this has the `read` but not `write` annotation, so this doesn't really track anything except return the value. It does indicate that both operations must be accounted for; however, I do not see why that needs to be explicitly stated.

The second comment regarding the type and the actual return type. This should really be set here rather than suggesting that another type may be used. If this is meant to be a standard then allowing users to mix and match types does not conform to the idea of a standard.
I'd probably just go with a u64 for simplicity.

(So I'd remove the last 2 comments and change the return type)

libs/frc20_abi/src/lib.sw
```diff
@@ -0,0 +1,22 @@
...
+    fn total_supply() -> U256;
+
+    /// Get the name of the token
+    /// Example (with trailing padding): "MY_TOKEN                                                        "
+    fn name() -> str[64];
```
Braqzen, line 14: Bytes support is gonna hit the Rust SDK soon. I'd use that type instead or perhaps the String type from this repo. cc @bitzoic ?

libs/frc20_abi/Forc.toml
```diff
@@ -0,0 +1,5 @@
+[project]
+authors = ["Fuel Labs <contact@fuel.sh>"]
+entry = "lib.sw"
+license = "Apache-2.0"
+name = "frc20_abi"
```
Braqzen, line 5: Interesting name, `frc`. Also, why 20? Are we following Ethereum for a specific reason or can we have some other convention?

libs/frc20_abi/src/lib.sw
nfurfaro: Thanks @Braqzen. Yeah, I'm removing the last 2 comments; while it is normal to specify this sort of thing in a token standard, they don't belong on a simple getter.
See [erc20](https://eips.ethereum.org/EIPS/eip-20#specification) for an example, ie `transfer` which specifies which event must be emitted.
I'd consider removing the comment about types. 
We discussed this at length on Slack (which is not to say there was 100% consensus), and a U256 allows for tokens with  extremely large supplies, and/or very high precision needs. The standard specifies the function signatures, but how the user implements storage details is up to them.

libs/frc20_abi/src/lib.sw
nfurfaro: Pretty sure `String`s can't be stored.

libs/frc20_abi/Forc.toml
pixelcircuits: yeah, reserving all the common numbers for parallels to Ethereum makes sense to me

libs/frc20_abi/src/lib.sw
pixelcircuits: I think we all decided fixed size strings is perfectly fine for now. Just pad out the name with trailing spaces and any front-end will know to trim it

pixelcircuits (approved):

PR #125, opened 2023-03-28
Queue

rostyslavtyshko:
## Type of change

- New feature

## Changes

Queue added

## Notes

- Note 1

## Related Issues

Closes #\<issue number\>


tests/src/queue/src/main.sw
nfurfaro, line 1: This is a perfect use case for in-language tests. They can be placed in the same src file as the Queue type, and then there's no need for the SDK at all. Another benefit is that this scrip[t can be refactored into 3 different tests, and a failure in one is easy to pinpoint. For reference, check out the tests in`std::bytes`.
If CI does not currently run `forc test` as one of its actions, we can copy over that action for the sway repo first as a separate PR.

libs/queue/src/main.sw
```diff
@@ -0,0 +1,39 @@
+library queue;
```
nfurfaro, line 1: See `std::bytes` for a reference; this needs docs added.

nfurfaro (commented):
This looks really good, and concise. I added a cou[p;le comments on docs and testing.

tests/src/queue/src/main.sw
rostyslavtyshko: Issue for the CI step https://github.com/FuelLabs/sway-libs/issues/135

libs/queue/src/main.sw
rostyslavtyshko: Added docs

tests/src/queue/src/main.sw
rostyslavtyshko: CI step added, tests added

rostyslavtyshko:
> Is this ready for review?

Yes it is

tests/src/queue/tests/mod.rs
```diff
@@ -0,0 +1,21 @@
...
+    use super::*;
+
+    #[tokio::test]
+    async fn runs_queue_test_script() {
+        let path_to_bin = "src/queue/out/debug/queue_test.bin";
+        let wallet = launch_provider_and_get_wallet().await;
+
+        let instance = TestQueue::new(wallet, path_to_bin);
+
+        let _result = instance.main().call().await;
```
Braqzen, line 19: This should `assert!(<return_value>)` otherwise this does not test anything aside from that we do not revert.

libs/Forc.toml
```diff
@@ -1,5 +1,6 @@
 [workspace]
 members = [
+  "queue",
```
Braqzen, line 3: I would move the queue to be in alphabetical order rather than at the top.

libs/queue/src/main.sw
```diff
@@ -0,0 +1,98 @@
+// The `Queue` type corresponds to the same data structure.
```
Braqzen, line 1: This _comment_ does not inform me of what a Queue type is or what the data structure may be. Moreover, it's a comment (uses 2 forward slashes) rather than a documentation comment (uses 3 forward slashes).
I would elaborate on what a queue is for those who have never heard of such a type.

libs/queue/src/main.sw
```diff
@@ -0,0 +1,98 @@
...
+        self.vec.is_empty()
+    }
+
+    /// Gets the number of elements in the `Queue`
+    pub fn len(self) -> u64 {
+        self.vec.len()
+    }
+
+    /// Enqueues a `Queue`
+    pub fn enqueue(ref mut self, item: T) {
```
Braqzen, line 29: I would add a section for the argument which explains what the `item` is.

libs/queue/src/main.sw
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    /// Dequeues a `Queue`
+    pub fn dequeue(ref mut self) -> Option<T> {
+        if self.vec.is_empty() {
+            Option::None
+        } else {
+            Option::Some(self.vec.remove(0))
+        }
+    }
```
Braqzen, line 40:
```suggestion
    /// Dequeues a `Queue`
    pub fn dequeue(ref mut self) -> Option<T> {
        if self.vec.is_empty() {
            return Option::None;
        }
        
        return Option::Some(self.vec.remove(0));
    }
```

Usually, it's better to structure code such that unnecessary nesting / indentation can be removed, this is particularly important when there are multiple levels of nested code.

libs/queue/src/main.sw
```diff
@@ -0,0 +1,98 @@
...
+    /// Gets the last element
+    pub fn peek(self) -> Option<T> {
+        self.vec.get(0)
+    }
+}
+
+#[test()]
+fn test_new_queue() {
+    let new_queue: Queue<u64> = Queue::new();
+    assert(new_queue.is_empty() == true);
```
Braqzen, line 51:
```suggestion
    assert(new_queue.is_empty());
```

Being explicit is useful (for example to catch regressions); however, it's cleaner to not equate boolean values via an equality expression. The assert function already checks if something is `true` so we don't need to check that here.

libs/queue/src/main.sw
```diff
@@ -0,0 +1,98 @@
...
+    assert(new_queue.len() == 0);
+}
+
+#[test()]
+fn test_enqueue() {
+    let mut queue_for_enqueue: Queue<u64> = Queue::new();
+    queue_for_enqueue.enqueue(1);
+    queue_for_enqueue.enqueue(2);
+    queue_for_enqueue.enqueue(3);
+    assert(queue_for_enqueue.len() == 3);
```
Braqzen, line 61: I would `enqueue` twice instead of 3 times too as 3 is unnecessary and after each addition I would assert that the length has increased by 1 otherwise this test proves that if you add 3 times then the length is 3 rather than after each addition the length increases by 1 because an item has been added.

libs/queue/src/main.sw
```diff
@@ -0,0 +1,98 @@
...
+}
+
+#[test()]
+fn test_dequeue() {
+    let mut queue_for_dequeue: Queue<u64> = Queue::new();
+    queue_for_dequeue.enqueue(1);
+    queue_for_dequeue.enqueue(2);
+    queue_for_dequeue.enqueue(3);
+    assert(queue_for_dequeue.dequeue().unwrap() == 1);
```
Braqzen, line 70:
```suggestion
    queue_for_dequeue.enqueue(3);
    assert(queue_for_dequeue.len() == 3);
    assert(queue_for_dequeue.dequeue().unwrap() == 1);
    assert(queue_for_dequeue.len() == 2);
```

Before you test your code you should assert that the state is what you believe it to be before you attempt to change it.

Similar comment to `enqueue` for checking the length

libs/queue/src/main.sw
```diff
@@ -0,0 +1,98 @@
...
+            Option::None
+        } else {
+            Option::Some(self.vec.remove(0))
+        }
+    }
+
+    /// Gets the last element
+    pub fn peek(self) -> Option<T> {
+        self.vec.get(0)
+    }
```
Braqzen, line 45: This states that it gets the last element but it gets the 0th index which is always the first element

tests/src/queue/src/main.sw
```diff
@@ -0,0 +1,45 @@
+script;
+
+use queue::Queue;
+use std::assert::assert;
```
Braqzen, line 4:
```suggestion
```

Assert is in the prelude so we can remove it

libs/queue/src/main.sw
```diff
@@ -0,0 +1,98 @@
+// The `Queue` type corresponds to the same data structure.
+library;
+
+use std::vec::Vec;
+
```
Braqzen, line 5:
```suggestion
```

The vector comes with the prelude therefore we do not need to import it ourselves.

tests/src/queue/src/main.sw
```diff
@@ -0,0 +1,45 @@
...
+
+fn main() -> bool {
+    let new_queue: Queue<u64> = Queue::new();
+    assert(new_queue.is_empty() == true);
```
Braqzen, line 8:
```suggestion
    assert(new_queue.is_empty());
```

tests/src/queue/src/main.sw
Braqzen, line 1: Looking at this, the script looks similar to the in-language tests. Is there a reason that this script needs to exist? Seems like it should only be tested in-language and the Rust tests should be entirely removed.

Braqzen (commented):
Looking at the code itself I'm wondering what the additional functionality of the queue is here as opposed to being a soft wrapper around a vector. I think this may need additional features to make it unique from the vector.

tests/src/queue/tests/mod.rs
rostyslavtyshko: Removed Rust tests

tests/src/queue/src/main.sw
rostyslavtyshko: Removed Rust tests

tests/src/queue/src/main.sw
rostyslavtyshko: Removed Rust tests

tests/src/queue/src/main.sw
rostyslavtyshko: Good idea. Removed Rust tests

libs/queue/src/main.sw
rostyslavtyshko: That was meaning the last item in the queue, not the underlying `Vec`. Changed to more converntional head/tail terminology.

libs/queue/src/main.sw
rostyslavtyshko: Fixed as outlined, though IMHO returns that break the flow are worse that nested code, but that's proobably up to personal taste, at least here

libs/queue/src/main.sw
rostyslavtyshko: Removed

libs/Forc.toml
rostyslavtyshko: Moved

libs/queue/src/main.sw
rostyslavtyshko: Added module level docs.

libs/queue/src/main.sw
rostyslavtyshko: Done

libs/queue/src/main.sw
rostyslavtyshko: Added such section

libs/queue/src/main.sw
rostyslavtyshko: Added such checks

rostyslavtyshko:
> Looking at the code itself I'm wondering what the additional functionality of the queue is here as opposed to being a soft wrapper around a vector. I think this may need additional features to make it unique from the vector.

As any other `Queue` type in any other language first and foremost it guarantees and Queue data structure behaviour is maintained i.e. that by no means it was possible to add items somewhere in the middle, that one was getting items from specific place etc. Therefore the main purpose for existence is mainly to minimise useless mental overhead for developers and auditors.

libs/queue/src/main.sw
```diff
@@ -0,0 +1,101 @@
...
+    let new_queue: Queue<u64> = Queue::new();
+    assert(new_queue.is_empty());
+    assert(new_queue.len() == 0);
+}
+
+#[test()]
+fn test_enqueue() {
+    let mut queue_for_enqueue: Queue<u64> = Queue::new();
+    queue_for_enqueue.enqueue(1);
+    let queue_len = queue_for_enqueue.len();
```
Braqzen, line 59:
```suggestion
    let queue_len = queue_for_enqueue.len();
    assert(queue_len == 1);
```

rostyslavtyshko: Fixed

Braqzen (approved):

PR #130, opened 2023-04-10
Update Codeowners

Braqzen:
## Type of change

- Other (describe below)

## Changes

- Added team


.github/CODEOWNERS
```diff
@@ -1 +1 @@
-* @simonr0204
+* @braqzen @bitzoic
```
bitzoic, line 1:
```suggestion
* @nfurfaro 
```

bitzoic (changes requested):

.github/CODEOWNERS
Braqzen: https://github.com/FuelLabs/sway-applications/pull/619#discussion_r1161747499

K1-R1 (approved):

.github/CODEOWNERS
nfurfaro: If anyone else on team @FuelLabs/application-dev want notifications for the repo, I suggest we make the team the codeowner here as well (same as the app-dev repo).

Braqzen: I've only re-added myself in the apps repo because I didn't want others to get spammed but then the whole thing happened so I've suggested the same change here.
Either way, now that people should know what that file does I'm happy to be the only one added to it in both repos if they do not want to get spammed with review notifications by default.

PR #144, opened 2023-04-25
Deque

### LINKED ISSUE #134: Deque
### Motivation

Deque based on Vec

rostyslavtyshko:
## Type of change

- New feature
 `Deque` library

## Notes

- Note 1

## Related Issues

Closes #134


libs/deque/src/main.sw
```diff
@@ -0,0 +1,75 @@
...
+            back: Vec::with_capacity(capacity),
+        }
+    }
+
+    /// Appends an element to the back of the deque.
+    pub fn push_back(ref mut self, value: T) {
+        self.back.push(value);
+    }
+
+    /// Inserts an element at the front of the deque.
```
Braqzen, line 32:
```suggestion
    /// Inserts an element at the start of the deque.
```

Here too.

tests/src/deque/src/main.sw
```diff
@@ -0,0 +1,37 @@
+script;
+
+use deque::Deque;
+use std::assert::assert;
+
+fn main() -> bool {
+    let mut deque_push_front = Deque::new();
+    deque_push_front.push_front(1);
```
Braqzen, line 8: The `push_front()` test can be written as follows

1. Create a new Deque
2. Assert that the instance has no items
3. Push to the front once
4. Assert that the instance has increased in length by 1 and the item in the instance has the value that you have added
5. Push once more
6. Assert again that the length has increased by 1 and now assert that the first item and second item both have the correct values.

This ensures that each call to your function behaves as expected and there are no side effects after X number of calls.

We can be repeated for other test cases.

tests/src/deque/src/main.sw
```diff
@@ -0,0 +1,37 @@
...
+    deque_push_front.push_front(2);
+    deque_push_front.push_front(3);
+
+    assert(deque_push_front.pop_front().unwrap() == 3);
```
Braqzen, line 12: There are additional combinations for `push_front()`, `push_back()`, `pop_front()` & `pop_back` that should be tested here because popping affects the `front` and `back` fields depending on how the struct has been set up for the test.

libs/deque/src/main.sw
```diff
@@ -0,0 +1,75 @@
...
+    pub fn push_front(ref mut self, value: T) {
+        self.front.push(value);
+    }
+
+    pub fn pop_front(ref mut self) -> Option<T> {
+        if self.front.is_empty() {
+            if !self.back.is_empty() {
+                self.front.push(self.back.remove(0))
```
Braqzen, line 40:
```suggestion
               self.back.remove(0)
```

Is there a reason that we do not actually remove from the `front`? Here we remove the first item from the `back` but then we add it to the end of the `front` which effectively does not pop the item.

May need to run `forc fmt` here. and same below

libs/deque/src/main.sw
```diff
@@ -0,0 +1,75 @@
...
+    pub fn push_back(ref mut self, value: T) {
+        self.back.push(value);
+    }
+
+    /// Inserts an element at the front of the deque.
+    pub fn push_front(ref mut self, value: T) {
+        self.front.push(value);
+    }
+
+    pub fn pop_front(ref mut self) -> Option<T> {
```
Braqzen, line 37: We should document these methods

Braqzen (commented):
1. We should test the code via in-language tests rather than Rust code.
2. We do not need to import modules from the prelude e.g. the `assert` is automatically imported for us
3. I'm uncertain about this implementation which involves 2 fields. It does not behave as I expect, but I could be incorrect in my assumption. I expect something to track where the start / end is via an index and then using the index to determine where we insert / remove the item based on the method called. I would reconsider this implementation

libs/deque/src/main.sw
```diff
@@ -0,0 +1,75 @@
...
+
+    /// Creates a new, empty deque with the specified capacity.
+    pub fn with_capacity(capacity: u64) -> Self {
+        Self {
+            front: Vec::with_capacity(capacity),
+            back: Vec::with_capacity(capacity),
+        }
+    }
+
+    /// Appends an element to the back of the deque.
```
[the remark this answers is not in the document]
rostyslavtyshko: I guess it would be better to use conventional terminology (`back`, `front` etc), so that the users are comfortable with the library.

libs/deque/src/main.sw
rostyslavtyshko: Same argument as [here](https://github.com/FuelLabs/sway-libs/pull/144#discussion_r1182520989)

nfurfaro:
If the name of this lib follows Rust's conventions then disregard this comment, but I feel like we could make the lib name "dequeue" to follow along with the "queue" lib?
 The shorter "deque" only saves 2 chars...

rostyslavtyshko:
> 1. We should test the code via in-language tests rather than Rust code.
> 
>     2. We do not need to import modules from the prelude e.g. the `assert` is automatically imported for us
> 
>     3. I'm uncertain about this implementation which involves 2 fields. It does not behave as I expect, but I could be incorrect in my assumption. I expect something to track where the start / end is via an index and then using the index to determine where we insert / remove the item based on the method called. I would reconsider this implementation

1. Moved tests from Rust.
2. Not an issue since all the tests are now in-language.
3.

libs/deque/src/main.sw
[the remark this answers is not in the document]
Braqzen: Is there a reason to not use a similar implementation from Rust @rostyslavtyshko ?

bitzoic:
Closing due to inactivity

PR #147, opened 2023-04-27
Add test for cross-contract reentrancy

nfurfaro:
## Type of change

## Changes

- added a new test case
- added supporting new contract & abi
- added new functions on existing contracts & abis

## Notes

The test case added performs the following steps:

1. The `Attack` contract's `launch_thwarted_attack_3()` function is triggered by a contract call from the SDK.
2. The `Attack` contract calls into the `Target` contract's `cross_contract_reentrancy_denied()` function.
3. The `Target` contract calls back to the `Attack` contract's `evil_callback_4()` function.
4. The `Attack` contract calls into the `Attack Helper` contract's `attempt_cross_contract_reentrancy()` function.
5. (attack vector) The `Attack Helper` contract attempts to re-enter by calling into the `Target` contract's `cross_contract_reentrancy_denied()` function(which hasn't finished execution of the initial call from the `Attack` contract).
6.  The attack is thwarted by the `reentrancy_guard()` function in place.


tests/src/reentrancy/reentrancy_attack_helper_contract/Forc.toml
```diff
@@ -0,0 +1,9 @@
+[project]
+authors = ["   Nicholas Furfaro"]
```
Braqzen, line 2:
```suggestion
authors = ["Nicholas Furfaro"]
```

Braqzen (approved):
Seems fine but is there something preventing a 2 contract interaction over 3 contracts?
Contract A calls into Contract B (non-reentrant func)
Contract B calls into Contract A
Contract A calls back into Contract B (non-reentrant func)

pixelcircuits (approved):

tests/src/reentrancy/reentrancy_attack_helper_contract/Forc.toml
nfurfaro: fixed.

nfurfaro:
>Seems fine but is there something preventing a 2 contract interaction over 3 contracts?

@Braqzen I'm not quite sure what you're getting at here.
If the `reentrancy_guard()` function is _not_ used, given 2 Contracts `A` & `B`:
- `A` can call into `B` can call into `A` can call into `B` can call into `A` can call into `B` ... until the gas runs out.
- The compiler just warns us if the CEI (Checks, Effects, Interactions) pattern is violated. It doesn't strictly prevent reentrant function calls.

Braqzen:
> > Seems fine but is there something preventing a 2 contract interaction over 3 contracts?
> 
> @Braqzen I'm not quite sure what you're getting at here. If the `reentrancy_guard()` function is _not_ used, given 2 Contracts `A` & `B`:
> 
> * `A` can call into `B` can call into `A` can call into `B` can call into `A` can call into `B` ... until the gas runs out.
> * The compiler just warns us if the CEI (Checks, Effects, Interactions) pattern is violated. It doesn't strictly prevent reentrant function calls.

By "non-reentrant func" I meant to use the guard in that function, I should have been clearer, my mistake.

K1-R1 (approved):

PR #154, opened 2023-05-03
Add `ref mut` keyword for String Library function definitions

### LINKED ISSUE #153: Add `ref` keyword to `String` type
#### Description

After reporting that we can no longer use the underlying `Bytes` type [here](https://github.com/FuelLabs/sway/issues/4408), it has been reported that adding the `ref` keyword should fix this.

#### Motivation

Resolution to https://github.com/FuelLabs/sway/issues/4408 found.

bitzoic:
## Type of change

## Changes

- Added `ref mut` keyword to functions to use underlying type
- Uncommented blocked tests
- Updated README

## Notes

- While the issue of using `my_function(self)` still does not allow for modification of the underlying type, it was discussed that adding `ref mut` to the function argument will allow for this. This was brought up in the https://github.com/FuelLabs/sway/issues/4408 issue.

## Related Issues

Closes #153


libs/strings/string/src/lib.sw
Braqzen, line 1: The following methods do not mutate so we should remove the `mut` from them

- `as_vec()`
- `capacity()`
- `is_empty()`
- `len()`
- `nth()`

tests/src/string/src/main.sw
```diff
@@ -39,87 +39,90 @@ abi StringTest {
...
+        string2.push(NUMBER3);
+        string2.push(NUMBER4);
+        string2.push(NUMBER5);
+
+        assert(string2.len() == 3);
+        assert(string1.len() == 3);
+
+        string1.append(string2);
+
+        assert(string2.len() == 0);
```
Braqzen, line 57: Should `string2` give away its values or should `string1` only have those values added?

tests/src/string/src/main.sw
```diff
@@ -39,87 +39,90 @@ abi StringTest {
...
+        assert(bytes.len() == string.len());
+        assert(bytes.capacity() == string.capacity());
+        assert(bytes.get(1).unwrap() == string.nth(1).unwrap());
+
+        let result_string = string.pop().unwrap();
+        let result_bytes = bytes.pop().unwrap();
+        assert(result_bytes == result_string);
+        assert(bytes.len() == string.len());
+        assert(bytes.capacity() == string.capacity());
+        assert(bytes.get(0).unwrap() == string.nth(0).unwrap());
```
Braqzen, line 91: This does not directly test the `as_vec()` so I would remove it

tests/src/string/src/main.sw
```diff
@@ -130,116 +133,107 @@ impl StringTest for Contract {
...
 
-        // string.push(NUMBER0);
-        // assert(!string.is_empty());
+        string.push(NUMBER0);
+        assert(!string.is_empty());
 
-        // string.clear();
-        // assert(string.is_empty());
+        string.clear();
+        assert(string.is_empty());
```
Braqzen, line 163: This is excessive, I would remove this since we already test the addition and clearing above

tests/src/string/src/main.sw
```diff
@@ -130,116 +133,107 @@ impl StringTest for Contract {
...
     fn test_insert() {
         let mut string = String::new();
 
         assert(string.len() == 0);
 
-        // Uncomment when https://github.com/FuelLabs/sway/issues/4408 is resolved
-        // string.insert(NUMBER0, 0);
-        // assert(string.len() == 1);
-        // assert(string.nth(0).unwrap() == NUMBER0);
+        string.insert(NUMBER0, 0);
```
Braqzen, line 222: I would fill the string with a few values and then insert over them while asserting that they have changed after each insertion rather than mixing up inserting, pushing, asserting with various offsets

tests/src/string/src/main.sw
```diff
@@ -442,170 +428,161 @@ impl StringTest for Contract {
...
-        // assert(string.nth(1).unwrap() == NUMBER1);
-        // assert(string.nth(2).unwrap() == NUMBER2);
+        assert(string.nth(0).unwrap() == NUMBER0);
+        assert(string.nth(1).unwrap() == NUMBER1);
+        assert(string.nth(2).unwrap() == NUMBER2);
 
-        // string.set(0, NUMBER3);
-        // string.set(1, NUMBER4);
-        // string.set(2, NUMBER5);
+        string.set(0, NUMBER3);
```
Braqzen, line 498: Since we are checking the setting we should assert after each call otherwise it could be that after the call to `set` 3 times with any values could result in a false positive pass of this test because of the underlying implementation

tests/src/string/src/main.sw
bitzoic: Yes https://github.com/FuelLabs/sway/blob/bbe746dafa102dba3ca4a392f23548515e122343/sway-lib-std/src/bytes.sw#L640

K1-R1 (approved):

PR #168, opened 2023-06-15
Queue Documentation and Specification

### LINKED ISSUE #159: Add Queue Documentation and Specification
#### Description

The Queue Library has recently added to the sway-libs repository. There should be documentation as to how to use the library in a README.md file at the root directory of the library. There should also be included a SPECIFICATION.md for it's functions and implementation. This should be linked in the source directory README.md file.

#### Motivation

There has been a new library added without documentation.

brandonsurh:
## Type of change

- Documentation

## Changes

- Added README.md
- Added SPECIFICATION.md

## Notes

- Needs a Fuel Logo made for the Queue Library

## Related Issues

Closes #159


bitzoic (dismissed):
This is great! Thanks for contributing 🚀

libs/queue/README.md
```diff
@@ -0,0 +1,64 @@
+# Overview
+
+The Queue library provides an implementation of a queue data structure. A Queue is a linear structure which follows the First-In-First-Out (FIFO) principle. This means that the elements added first are the ones that get removed first.
```
Braqzen, line 3:
```suggestion
A Queue is a linear structure which follows the First-In-First-Out (FIFO) principle. This means that the elements added first are the ones that get removed first.
```

The sentence seems redundant here

libs/queue/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+The `Queue<T>` is a structure with a generic type parameter T. The elements in the queue are stored as a `Vec<T>`. The data structure is created empty and is growable.
```
Braqzen, line 5:
```suggestion
The `Queue<T>` is a structure with a generic type parameter `T`. The elements in the queue are stored as a `Vec<T>`. The data structure is created empty and is growable.
```

The highlight is for consistency.

In terms of the content I would entirely remove the sentence or in-cooperate the use of the generic into some other place.
The reason is that as a user of the library you care about its utility rather than its implementation.
That internal representation would typically be encapsulated from the user similar to how an API informs you of how to use it rather than what it does under the hood (this also allows you to re-implement it without having to update docs here).

On the other hand, from the perspective of a developer you'd want to inform the reader of this kind of information. The place for that would be the specification rather than the more general "introductory" document which is the readme

libs/queue/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+For more information please see the [specification](./SPECIFICATION.md).
+
+> **Note** The Queue<T> type in this library is different from built-in arrays and vectors due to its unique characteristics, especially its operation in FIFO manner.
```
Braqzen, line 9: I would also remove this note as it does not inform the user of anything new / add value. It has been stated in the first sentence that it's a structure that follows FIFO and here it re-iterates that and points out that it's not a _different_ data structure - which is not something that someone would automatically think e.g. when thinking of an unsigned integer and a signed integer you may think about the commonality of them being integers; nevertheless, they are different and not interchangeable

libs/queue/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Known Issues
+
+Currently, the Queue<T> type does not provide any way to iterate through its elements.
```
Braqzen, line 13: I'd be weary of listing items that may be feature requests as opposed to issues.

For example, if the queue usually provides the feature for iteration but under some concrete type for `T` the iteration breaks then that is a known issue.
On the other hand, if the feature is entirely missing then that's not an issue, it's a missing feature.

Moreover, I would consider whether there are additional complications with the implementation.

libs/queue/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Known Issues
+
+Currently, the Queue<T> type does not provide any way to iterate through its elements. 
+
+# Using the Library
+
+## Getting Started
+
+To use the Queue library, you need to add it to the Forc.toml file and then import it into your Sway project. For more details on how to add Sway-libs as a dependency to your Forc.toml, refer to the [README.md](../../../README.md).
```
Braqzen, line 19: This is a broken README link. I recommend checking out where the file has been moved to and updating it - probably submit a PR to update other libraries if this file has been copied and pasted from another library

libs/queue/SPECIFICATION.md
```diff
@@ -0,0 +1,35 @@
...
+
+Returns a boolean indicating whether the length of the `Queue` is zero. 
+
+### `len()`
+
+Returns the number of elements in the `Queue`
+
+### `enqueue()`
+
+Adds the passed element to the `Queue`
```
Braqzen, line 27:
```suggestion
Adds the item into the `Queue`
```

Instead of "item" I guess "element" could be used too - we should be consistent with 1 or the other though

libs/queue/SPECIFICATION.md
```diff
@@ -0,0 +1,35 @@
...
+
+### `dequeue()`
+
+Removes an element to the `Queue`
```
Braqzen, line 31:
```suggestion
Removes an element from the `Queue`
```

Gonna leave "element" here instead of "item" above

libs/queue/SPECIFICATION.md
```diff
@@ -0,0 +1,35 @@
...
+
+### `peek()`
+
+Returns the head of the `Queue`
```
Braqzen, line 35: What does it mean to "return" the "head". The specification is the place for in depth documentation

libs/queue/SPECIFICATION.md
```diff
@@ -0,0 +1,35 @@
+# Overview
+
+This document provides an overview of the Queue library.
+
+It outlines the use cases, i.e. specification, and describes how to implement the library.
+
+## Use Cases
+
+The Queue library can be used as a traditional queue or first in first out data structure.
```
Braqzen, line 9: I don't know what this sentence is trying to tell me.

Should I assume that there are non-traditional queues, which is the traditional one? The end of the sentence mentions the FIFO context which is fine but this is the "use case" section and this does not describe a use case - it describes how the implementation functions.

I think the confusion is on the terminology of "use case".
A use case for a business could be to retrieve the first data packet that it receives i.e. FIFO (UDP in networking for example). The queue then meets that _requirement_ because it provides the functionality and thus the overall "use case" for the business is satisfied.

libs/queue/SPECIFICATION.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Public Functions
+
+### `new()`
+
+Creates a new instance of a `Queue`.
```
Braqzen, line 15:
```suggestion
- Creates a new instance of a `Queue`. 
```

I would add bullet points to all of these to make them more readable (especially when more items are added) and consistent across larger libraries (that probably do this / ought to if they are not).

Braqzen (commented):
Just some comments for improving the docs

libs/queue/README.md
brandonsurh: Good suggestion. Will make some adjustments.

libs/queue/README.md
brandonsurh: I will be making some adjustments to make sure the SPECIFICATION.md has new information that can't be found on the README.md. Thank you for pointing that one out. As for the suggestion about stating that Queues are different from the built-in arrays and vectors, are you recommending that I remove that portion of the note? I would think that it's worth mentioning that it is different than the built-in data structures but would like to know your thoughts.

libs/queue/README.md
brandonsurh: Got it, will remove it from Known Issues.

libs/queue/SPECIFICATION.md
brandonsurh: Got it, I will commit a rewritten version of this.

brandonsurh:
Made changes in regard to your suggestions. Let me know what you think. Thanks!

libs/queue/README.md
Braqzen: I would remove it.
I don't see why someone would assume that two different things are the same - even if they are ignorant.

If someone wants to use the library then it's likely that they already have an idea of what it is; moreover, if they do not know then showing them how to use the library should be sufficient. If you do not show them how to use a queue as an array (because it may not work that way) then they should not assume that they can use it that way.

As another example, if you want to purchase a car do you expect the salesperson to say "by the way, it does not fly because it's not a plane"?
There's no reason to state that information.

libs/queue/README.md
```diff
@@ -1,22 +1,16 @@
...
-## Known Issues
-
-Currently, the Queue<T> type does not provide any way to iterate through its elements. 
-
 # Using the Library
 
 ## Getting Started
 
-To use the Queue library, you need to add it to the Forc.toml file and then import it into your Sway project. For more details on how to add Sway-libs as a dependency to your Forc.toml, refer to the [README.md](../../../README.md).
+To use the Queue library, you need to add it to the Forc.toml file and then import it into your Sway project. For more details on how to add Sway-libs as a dependency to your Forc.toml, refer to the [README.md](./README.md).
```
Braqzen, line 13: This readme link points back to itself. It should point to the one in the root of the repo because that is where the description for how to use a library is documented

libs/queue/SPECIFICATION.md
```diff
@@ -4,32 +4,34 @@ This document provides an overview of the Queue library.
...
 
 ### `len()`
 
-Returns the number of elements in the `Queue`
+- Returns the number of elements in the `Queue`
 
 ### `enqueue()`
 
-Adds the passed element to the `Queue`
+- Adds the item into the `Queue`
```
Braqzen, line 29: As per my previous comment, if you're going to go with the suggestion of an `item` then every instance of the word `element` in the readme and spec should also be updated otherwise we should consistently use `element`

libs/queue/SPECIFICATION.md
```diff
@@ -4,32 +4,34 @@ This document provides an overview of the Queue library.
 
 It outlines the use cases, i.e. specification, and describes how to implement the library.
 
+The `Queue<T>` is a linear structure with a generic type parameter `T`. The elements in the queue are stored as a `Vec<T>`. Queues are growable, operate in a First-In-First-Out (FIFO) manner, and are created empty.
+
 ## Use Cases
 
-The Queue library can be used as a traditional queue or first in first out data structure.
+The Queue library can be used First-In-First-Out data structure when you want to get items out in the order that they are added.
```
Braqzen, line 11: This is basically repeating the first line of the readme.

You could probably expand on / simplify the part about retrieving items in the order that they are added.

libs/queue/SPECIFICATION.md
```diff
@@ -4,32 +4,34 @@ This document provides an overview of the Queue library.
 
 It outlines the use cases, i.e. specification, and describes how to implement the library.
 
+The `Queue<T>` is a linear structure with a generic type parameter `T`. The elements in the queue are stored as a `Vec<T>`. Queues are growable, operate in a First-In-First-Out (FIFO) manner, and are created empty.
```
Braqzen, line 7: This is a paragraph that really consists of 3 bullet points strung together. I would consider adjusting the format and presenting some heading / title to indicate what this is.

The overview begins by introducing the reader to the document itself and then suddenly into facts about the library. That's a large shift which can be eased by an appropriate grouping (heading / category etc.)

libs/queue/README.md
brandonsurh: I see your point with that. I'll take it out.

brandonsurh:
Made changes as per request. Let me know what you think or if there is anything I missed. 

My apologies for the emails about pushes. I didn't realize the GitHub file editor's "Commit Changes" button was spamming you guys.

Braqzen (approved):
I'd still make some changes but I've already commented a bunch of times so approving

PR #171, opened 2023-06-28
Refactor to use `From` trait for `String`

### LINKED ISSUE #170: Update `String` to use `From` trait over special functions
#### Description

The current implementation of `String` uses special functions for conversion between types. This should instead use the `From` trait.

#### Motivation

Multiple `From` implementations are now functioning as intended.

bitzoic:
## Type of change

## Changes

- Uses `From<Vec>` instead of `as_vec` and `from_utf8`
- Uses `From<raw_slice>` instead of `from_raw_slice` 

## Notes

- This is a breaking change

## Related Issues

Closes #170

### DIFF
--- a/libs/strings/string/src/lib.sw
+++ b/libs/strings/string/src/lib.sw
@@ -7,11 +7,6 @@ pub struct String {
 }
 
 impl String {
-    /// Returns a `Vec<u8>` of the bytes stored for the `String`.
-    pub fn as_vec(self) -> Vec<u8> {
-        self.bytes.into_vec_u8()
-    }
-
     /// Gets the amount of memory on the heap allocated to the `String`.
     pub fn capacity(self) -> u64 {
         self.bytes.capacity()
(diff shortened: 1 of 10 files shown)


libs/strings/string/src/lib.sw
```diff
@@ -173,17 +149,33 @@ impl String {
...
+        asm(ptr: (self.bytes.buf.ptr(), self.bytes.len)) { ptr: raw_slice }
+    }
+}
+
+impl From<Vec<u8>> for String {
+    /// Converts a vector of bytes to a `String`.
+    ///
+    /// # Arguments
+    ///
+    /// * `bytes` - The vector of `u8` bytes which will be converted into a `String`.
```
K1-R1, line 169: This doesn't match the `from` function below

K1-R1 (changes requested):

libs/strings/string/src/lib.sw
```diff
@@ -173,17 +149,33 @@ impl String {
...
-//             bytes
-//         }
-//     }
-
-//     fn into(self) -> raw_slice {
-//         asm(ptr: (self.bytes.buf.ptr(), self.bytes.len)) { ptr: raw_slice }
-//     }
-// }
+impl From<raw_slice> for String {
+    fn from(slice: raw_slice) -> String {
```
Braqzen, line 153:
```suggestion
    fn from(slice: raw_slice) -> Self {
```

libs/strings/string/src/lib.sw
```diff
@@ -173,17 +149,33 @@ impl String {
...
+        Self {
+            bytes: Bytes::from_raw_slice(slice),
```
Braqzen, line 155: I assume this will be updated to use the `from` on `Bytes`

bitzoic: Yes a new release is needed.

K1-R1 (approved):

PR #180, opened 2023-08-08
Update ownership to follow new standards

bitzoic:
## Type of change

- Documentation

## Changes

- Updated to follow the SRC-2 inline docs standard.
- Now imports and uses the SRC-5 Ownership standard.


libs/ownership/src/ownable.sw
```diff
@@ -1,53 +1,103 @@
...
-    owner: State,
-}
+use src_5::{Ownership, State};
 
 impl Ownership {
     /// Returns the `Ownership` struct in the `Uninitalized` state.
+    ///
+    /// # Returns
+    ///
+    /// * [Ownership] - The struct in the `Uninitalized` state.
```
Braqzen, line 16: Given the first line do we need to reinstate this?

Braqzen: Same comment for a few additions below

PR #199, opened 2023-10-23
Introduce the Pausable Library

### LINKED ISSUE #198: Pausable Library
### Description 

Certain apps implement a pause function when initially deploying to give it "training wheels". This functionality should be encapsulated into a library.

### Motivation

To enable an emergency stop function.

bitzoic:
## Type of change

- New Library

## Changes

- Adds the Pausable Library to the sway-libs repo

## Notes

- This library is completely stateless and relies on a pre-defined sub-id to mint and burn an asset that represents a state. This is being called "packet-oriented design" and saves users 20% gas in my tests as opposed to storing a `bool` in storage. It also does not require that the developer pass a `StorageKey` for the library to properly function.

## Related Issues

Closes #198


libs/pausable/src/lib.sw
```diff
@@ -0,0 +1,181 @@
...
+    /// # Examples
+    ///
+    /// ```sway
+    /// use pausable::Pausable;
+    ///
+    /// fn foo(contract: ContractId) {
+    ///     let pausable_abi = abi(Pauseable, contract);
+    ///     pausable_abi.pause();
+    ///     assert(pausable_abi.is_paused() == true);
+    /// }
```
Braqzen, line 32:
```suggestion
    /// fn foo(contract_id: ContractId) {
    ///     let pausable_abi = abi(Pauseable, contract_id);
    ///     pausable_abi.pause();
    ///     assert(pausable_abi.is_paused());
    /// }
```

Keyword + CQ

libs/pausable/src/lib.sw
```diff
@@ -0,0 +1,181 @@
...
+///     assert(_is_paused() == false);
+/// }
+/// ```
+pub fn _unpause() {
+    if balance() == 1 {
+        burn(PAUSABLE, 1);
+    }
+}
+
+/// Returns whether the contract is in the paused or unpaused state.
```
Braqzen, line 111:
```suggestion
/// Returns whether the contract is in the paused state.
```

libs/pausable/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+# Using the Library
+
+## Getting Started
+
+In order to use the Pausable library it must be added to the `Forc.toml` file and then imported into your Sway project. To add Pausable as a dependency to the `Forc.toml` file in your project please see the [README.md](../../README.md).
+
+You may import the Pausable library's functionalities like so:
+
+```rust
```
Braqzen, line 26:
```suggestion
```sway
```

Braqzen (commented):
Outside of global search and replace knits in this whole PR there's only 1 point to make.

This is an interesting design mechanism but its design is flawed. There is a comment to use admin restrictions nevertheless it allows for spamming of an asset if somewhere where to go wrong and have the balance altered. This may affect the market.

Moreover, it's really a hack to save costs rather than a solid design. I don't believe in promoting hacks. If this wasn't a security based library then I'd have less of an issue with a hack and some asset potentially being nuked in value so I'd go back to statefulness

bitzoic:
> This is an interesting design mechanism but its design is flawed. There is a comment to use admin restrictions nevertheless it allows for spamming of an asset if somewhere where to go wrong and have the balance altered. This may affect the market.

Your comment makes no sense. The design is not flawed. Admin restrictions only allow a SINGLE user to pause rather than anyone. There cannot be spamming of an asset, when minting may only happen in the case of NO tokens. There is NO market involved.

> Moreover, it's really a hack to save costs rather than a solid design. I don't believe in promoting hacks. If this wasn't a security based library then I'd have less of an issue with a hack and some asset potentially being nuked in value so I'd go back to statefulness

This is NOT a hack, it is leveraging UTXO design. This was discussed a number of times as packet packet-oriented design with the use of Sub Ids. Again, there is NO value.

Braqzen:
> > This is an interesting design mechanism but its design is flawed. There is a comment to use admin restrictions nevertheless it allows for spamming of an asset if somewhere where to go wrong and have the balance altered. This may affect the market.
> 
> Your comment makes no sense. The design is not flawed. Admin restrictions only allow a SINGLE user to pause rather than anyone. There cannot be spamming of an asset, when minting may only happen in the case of NO tokens. There is NO market involved.
> 
> > Moreover, it's really a hack to save costs rather than a solid design. I don't believe in promoting hacks. If this wasn't a security based library then I'd have less of an issue with a hack and some asset potentially being nuked in value so I'd go back to statefulness
> 
> This is NOT a hack, it is leveraging UTXO design. This was discussed a number of times as packet packet-oriented design with the use of Sub Ids. Again, there is NO value.

Pause -> mint -> somehow withdraw the balance -> in "paused" but actually unpaused state. Design flaw.
Again, you're wrong. It is a hack. You're minting outside the purpose of a mint. If you could create and destroy something that wouldn't have any value, equivalent to a bool, that would be fine.

bitzoic:
> > > This is an interesting design mechanism but its design is flawed. There is a comment to use admin restrictions nevertheless it allows for spamming of an asset if somewhere where to go wrong and have the balance altered. This may affect the market.
> > 
> > 
> > Your comment makes no sense. The design is not flawed. Admin restrictions only allow a SINGLE user to pause rather than anyone. There cannot be spamming of an asset, when minting may only happen in the case of NO tokens. There is NO market involved.
> > > Moreover, it's really a hack to save costs rather than a solid design. I don't believe in promoting hacks. If this wasn't a security based library then I'd have less of an issue with a hack and some asset potentially being nuked in value so I'd go back to statefulness
> > 
> > 
> > This is NOT a hack, it is leveraging UTXO design. This was discussed a number of times as packet packet-oriented design with the use of Sub Ids. Again, there is NO value.
> 
> Pause -> mint -> somehow withdraw the balance -> in "paused" but actually unpaused state. Design flaw. Again, you're wrong. It is a hack. You're minting outside the purpose of a mint. If you could create and destroy something that wouldn't have any value, equivalent to a bool, that would be fine.

"somehow withdraw the balance" is an impossible hypothetical situation considering the contract owns the minted asset. Again, the asset has no value.

bitzoic:
Closed in favor of https://github.com/FuelLabs/sway-libs/pull/200

PR #200, opened 2023-10-27
Introduce the Pausable Library

bitzoic:
## Type of change

- New Library

## Changes

- Introduces the pausable library to the repo

## Notes

- The original design used a packet-oriented approach. However due to changes in gas outlined [here](https://github.com/FuelLabs/fuel-core/pull/1408#issuecomment-1771738623), a storage approach has instead been selected as mint/burn opcodes have become significantly more expensive.

## Related Issues

Closes [#199 ](https://github.com/FuelLabs/sway-libs/issues/198)


libs/pausable/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+In order to use the Pausable library it must be added to the `Forc.toml` file and then imported into your Sway project. To add Pausable as a dependency to the `Forc.toml` file in your project please see the [README.md](../../README.md).
+
+You may import the Pausable library's functionalities like so:
+
+```sway
+use pausable::*;
+```
+
+Be sure to add the storage block bellow to your contract which enables the library.
```
Braqzen, line 28:
```suggestion
Be sure to add the storage block below to your contract which enables the library.
```

How about we write directly to some storage slot using read/write? No need for them to use a storage field then

tests/src/pausable/src/main.sw
```diff
@@ -0,0 +1,101 @@
...
+    fn test_require_not_paused() {
+        require_not_paused(storage.paused);
+    }
+}
+
+#[test]
+fn test_is_paused() {
+    let pausable_abi = abi(Pausable, CONTRACT_ID);
+
+    assert(pausable_abi.is_paused() == false);
```
Braqzen, line 49:
```suggestion
    assert(!pausable_abi.is_paused());
```

por favor

libs/pausable/README.md
bitzoic: We used to do this before we had `StorageKey`s. Might not be the worst thing to implement again.

K1-R1 (approved):

PR #204, opened 2023-11-16
Refactor Ownership Library for better UX

bitzoic:
## Type of change

## Changes

- Swaps the need for declaring ownership in storage for a pre-defined `StorageKey` to store storage. This results in the following improvements in developer UX as the use of the library was previously very verbose.

Before:
```sway
storage {
     owner: Ownership = Ownership::initialized(Identity::Address(Address::from(ZERO_B256))),
}

fn foo() {
     let owner = storage.owner.owner();
     storage.owner.only_owner();
     storage.owner.renounce_ownership();
}
```
After:
```sway
fn foo() {
     let owner = _owner();
     only_owner();
     renounce_ownership();
}
```

The downside to this approach is the inability to set the owner during compile time and a constructor must be implemented.

### DIFF
--- a/libs/ownership/README.md
+++ b/libs/ownership/README.md
@@ -25,7 +25,7 @@ You may import the Ownership library's functionalities like so:
 use ownership::*;
 ```
 
-Once imported, the Ownership library's functions should be available. To use them, be sure to initialize the owner for your contract by calling the `initialize_ownership()` function in your own constructor.
+Once imported, the Ownership library's functions will be available. To use them initialize the owner for your contract by calling the `initialize_ownership()` function in your own constructor method.
 
 ```sway
 #[storage(read, write)]
@@ -65,6 +65,6 @@ impl SRC5 for Contract {
 }
 ```
 
-> **NOTE** A constructor must be implemented to initialize the owner.
+> **NOTE** A constructor method must be implemented to initialize the owner.
 
 For more information please see the [specification](./SPECIFICATION.md).
--- a/libs/ownership/README.md
+++ b/libs/ownership/README.md
@@ -7,7 +7,7 @@
 
 # Overview
 
-The Ownership library provides a way to block users other than a single "owner" or "admin" from calling functions. Ownership is often used when needing administrative calls on a contract.
+The Ownership library provides a way to block users other than a single "owner" from calling functions. Ownership is often used when needing administrative calls on a contract.
 
 For more information please see the [specification](./SPECIFICATION.md).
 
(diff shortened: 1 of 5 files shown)


Braqzen (commented):
Before I review the changes themselves can you elaborate on the technical need for this change?

To me it seems like a subjective change in UX.
I think that having the ability to set data at compile time is important and useful. At the same time being able to delay that via a constructor may be useful as well.

Without deeper consideration at this stage I'll await your response and instead superficially propose the idea of having 2 ownership standards. One for configuration and one for construction. That has trade-offs as well but I don't think that's as much of an issue as providing one UX over the other.

bitzoic:
> Before I review the changes themselves can you elaborate on the technical need for this change?
> 
> To me it seems like a subjective change in UX. I think that having the ability to set data at compile time is important and useful. At the same time being able to delay that via a constructor may be useful as well.
> 
> Without deeper consideration at this stage I'll await your response and instead superficially propose the idea of having 2 ownership standards. One for configuration and one for construction. That has trade-offs as well but I don't think that's as much of an issue as providing one UX over the other.

This is something I considered, but there are a few caveats to this. 
1. In the future we expect to have the ability to import storage from libraries, thus enabling the ability to set the owner at compile time
2. With most developers coming from Solidity, a constructor is something they are already accustomed to
3. If a developer chooses to set the owner at compile time, they can do so using just the [SRC-5 standard](https://github.com/FuelLabs/sway-standards/blob/master/examples/src_5/initialized_example/src/initialized_example.sw). This library is intended for convenience and is by no means the only way to implement the standard

I believe the proposed changes make Sway code significantly more readable and less verbose. Having to call `storage.owner.my_function()` for absolutely everything simply doesn't make sense compared to just the function itself. The trade off being well worth the minor inconvenience of a constructor in turn for easier auditability.

libs/ownership/README.md
```diff
@@ -7,61 +7,64 @@
...
-}
-```
-
-2. Leave the owner uninitialized and call the `set_ownership()` function in your own constructor.
-
-```rust
-storage {
-    owner: Ownership = Ownership::uninitialized(),
-}
+Once imported, the Ownership library's functions should be available. To use them, be sure to initialize the owner for your contract by calling the `initialize_ownership()` function in your own constructor.
```
K1-R1, line 28:
```suggestion
Once imported, the Ownership library's functions will be available. To use them initialize the owner for your contract by calling the `initialize_ownership()` function in your own constructor method.
```

libs/ownership/README.md
```diff
@@ -7,61 +7,64 @@
...
+
+impl SRC5 for Contract {
+    #[storage(read)]
+    fn owner() -> State {
+        _owner()
+    }
+}
+```
+
+> **NOTE** A constructor must be implemented to initialize the owner.
```
K1-R1, line 68:
```suggestion
> **NOTE** A constructor method must be implemented to initialize the owner.
```

K1-R1 (approved):
Big fan of the UX improvements here. Just a couple of comments for the README, but they're only suggestions

Braqzen (approved):
<3

PR #206, opened 2023-11-23
Introduce Admin Library

bitzoic:
## Type of change

- New feature

## Changes

- Introduces the admin library which has the following functionality:
     - `add_admin()`
     - `revoke_admin()`
     - `is_admin()`
     - `only_admin()`
     - `only_owner_or_admin()`
- This library allows for multiple administrators rather than a single user.

## Notes

- This library is built atop the ownership library where the owner is the only user who may add administrators.

### DIFF
--- a/libs/admin/src/admin.sw
+++ b/libs/admin/src/admin.sw
@@ -1,0 +1,181 @@
+library;
+
+mod errors;
+
+use errors::AccessError;
+use ownership::{_owner, only_owner};
+use src_5::{State};
+use std::{
+    auth::msg_sender,
+    storage::storage_api::clear,
+};
+
+/// Sets a new administrator.
+///
+/// # Arguments
+///
+/// * `new_admin`: [Identity] = The `Identity` which is to recieve administrator status.
+///
+/// # Reverts
+///
+/// * When the caller is not the contract owner.
+///
+/// # Number of Storage Accesses
+///
+/// * Reads: `1`
+/// * Writes: `1`
+///
+/// # Examples
+///
+/// ```sway
+/// use admin::{add_admin, is_admin};
+///
+/// fn foo(new_admin: Identity) {
+///     add_admin(new_admin);
+///     assert(is_admin(new_admin));
+/// }
+/// ```
+#[storage(read, write)]
+pub fn add_admin(new_admin: Identity) {
+    only_owner();
+
+    let admin_value = match new_admin {
+        Identity::Address(addr) => addr.value,
+        Identity::ContractId(contr) => contr.value,
+    };
+    let admin_key = StorageKey::<Identity>::new(admin_value, 0, admin_value);
+
+    admin_key.write(new_admin);
+}
+
+/// Removes an administrator.
+///
+/// # Arguments
+///
+/// * `old_admin`: [Identity] = The `Identity` which the administrator status is to be removed.
+///
+/// # Reverts
+///
+/// * When the caller is not the contract owner.
+///
+/// # Number of Storage Accesses
+///
+/// * Reads: `1`
+/// * Writes: `1`
+///
+/// # Examples
+///
+/// ```sway
+/// use admin::{remove_admin, is_admin};
+///
+/// fn foo(old_admin: Identity) {
+///     remove_admin(old_admin);
+///     assert(!is_admin(old_admin));
+/// }
+/// ```
+#[storage(read, write)]
+pub fn remove_admin(old_admin: Identity) {
+    only_owner();
+
+    let admin_value = match old_admin {
+        Identity::Address(addr) => addr.value,
+        Identity::ContractId(contr) => contr.value,
+    };    
+    let admin_key = StorageKey::<Identity>::new(admin_value, 0, admin_value);
+
+    // TODO: Update to use StorageKey::clear() on next release
+    // https://github.com/FuelLabs/sway/pull/5284
+    let _ = clear::<Identity>(admin_key.slot, admin_key.offset);
+}
+
+/// Returns whether `admin` is an administrator.
+///
+/// # Arguments
+///
+/// * `admin`: [Identity] = The `Identity` of which to check the administrator status.
+///
+/// # Returns
+///
+/// * [bool] - `true` if the `admin` is an administrator, otherwise `false`.
+///
+/// # Number of Storage Accesses
+///
+/// * Reads: `1`
+///
+/// # Examples
+///
+/// ```sway
+/// use admin::{is_admin};
+///
+/// fn foo(admin: Identity) {
+///     assert(is_admin(admin));
+/// }
+/// ```
+#[storage(read)]
+pub fn is_admin(admin: Identity) -> bool {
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -37,7 +37,8 @@ These libraries contain helper functions and other tools valuable to blockchain
 
 #### Access Control and Security
 
-- [Ownership](./libs/ownership/) is used to apply restrictions on functions such that only a single user may call them.
+- [Ownership](./libs/ownership/) is used to apply restrictions on functions such that only a **single** user may call them.
+- [Admin](./libs/admin/) is used to apply restrictions on functions such that only a select few users may call them.
```
Braqzen, line 41:
```suggestion
- [Admin](./libs/admin/) is used to apply restrictions on functions such that only a select few users may call them.
- [Ownership](./libs/ownership/) is used to apply restrictions on functions such that only a **single** user may call them.
```

tests/src/admin/tests/functions/add_admin.rs
```diff
@@ -0,0 +1,59 @@
+use crate::admin::tests::utils::{
+    abi_calls::{add_admin, is_admin, set_ownership},
+    test_helpers::setup,
+};
+use fuels::types::Identity;
+
+mod success {
```
Braqzen, line 7: I'd convert to references instead of cloning but not really an issue

tests/src/admin/tests/utils/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+    types::Identity,
+};
+
+// Load abi from json
+abigen!(Contract(
+    name = "AdminLib",
+    abi = "src/admin/out/debug/admin_test-abi.json"
+));
+
+pub struct Metadata {
```
Braqzen, line 16:
```suggestion
pub struct User {
```

tests/src/admin/tests/utils/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+    pub contract: AdminLib<WalletUnlocked>,
+    pub wallet: WalletUnlocked,
+}
+
+pub mod abi_calls {
```
Braqzen, line 21:
```suggestion
pub mod interface {
```

tests/src/admin/tests/utils/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+        };
+
+        let admin1 = Metadata {
+            contract: AdminLib::new(id.clone(), wallet3.clone()),
+            wallet: wallet3.clone(),
+        };
+
+        let admin2 = Metadata {
+            contract: AdminLib::new(id.clone(), wallet4.clone()),
+            wallet: wallet3.clone(),
```
Braqzen, line 138: I'd run `cargo clippy --all-features --all-targets -- -D warnings` to see the warnings it throws. I would also add it to CI if it's not there.

libs/admin/src/admin.sw
```diff
@@ -0,0 +1,180 @@
...
+use errors::AdminError;
+use ownership::{_owner, only_owner};
+use src_5::State;
+use std::{auth::msg_sender, storage::storage_api::clear,};
+
+// Sets a new administrator.
+///
+/// # Arguments
+///
+/// * `new_admin`: [Identity] = The `Identity` which is to recieve administrator status.
```
Braqzen, line 14: Was this always an `=` or did the standard change a little?

```suggestion
/// * `new_admin`: [Identity] = The `Identity` which is to receive administrator status.
```

libs/admin/src/admin.sw
```diff
@@ -0,0 +1,180 @@
...
+        Identity::ContractId(contr) => contr.value,
+    };
+    let admin_key = StorageKey::<Identity>::new(admin_value, 0, admin_value);
+    match admin_key.try_read() {
+        Some(identity) => {
+            admin == identity
+        },
+        None => {
+            false
+        },
```
Braqzen, line 120:
```suggestion
        Some(identity) => admin == identity,
        None => false,
```

How about something like this? May need slight syntax adjustment

libs/admin/README.md
```diff
@@ -0,0 +1,86 @@
...
+#[storage(read)]
+fn check_if_admin(admin: Identity) {
+    let status = is_admin(admin);
+    assert(status);
+}
+```
+
+## Integrating the Admin Library into the Ownership Library
+
+To implement the Ownership library with the Admin library, be sure to add set a contract owner for your contract. The following demonstrates the integration of the Ownership library with the Admin library.
```
Braqzen, line 68:
```suggestion
To implement the Ownership library with the Admin library, be sure to set a contract owner for your contract. The following demonstrates the integration of the Ownership library with the Admin library.
```

Braqzen (commented):
As stated, this is an extension of ownership. It's a nice iterative improvement but this will be deprecated when role based access control is implemented.

README.md
bitzoic: It doesn't make any sense to list an extension before it's dependency.

tests/src/admin/tests/utils/mod.rs
bitzoic: This is not a user as it contains a contract.

tests/src/admin/tests/utils/mod.rs
bitzoic: I am leaving this as is. It is the same structure followed in other tests across 3 repositories. `abi_calls` is clear, concise, and quickly understandable.

libs/admin/src/admin.sw
bitzoic: This is the formatter's work

README.md
Braqzen: Keeping things organized by lexicographical ordering. No one except the author is thinking about dependency ordering here. From the user's perspective, and UX, it makes sense.

tests/src/admin/tests/utils/mod.rs
Braqzen: It belongs to the user though, no?

Either way, `Metadata` is a broad and vague name.

tests/src/admin/tests/utils/mod.rs
Braqzen: That's fine.

I've pointed out that `interface` is what we generally use and is the canonical term for a reason.

libs/admin/src/admin.sw
Braqzen: oh, so the formatter enforces this now? that's strange, some time ago we could have my suggestion or you could wrap it yourself and then it would format it your way

Braqzen:
> > As stated, this is an extension of ownership. It's a nice iterative improvement but this will be deprecated when role based access control is implemented.
> 
> This is the "role based access control" implementation.

oh no lol. this isn't role based permissions.
a role based access control allows a user to create roles and assign them to users.

this has a single hard coded role with functionality specifically for it so my original comment stands.

Braqzen (approved):

K1-R1 (approved):

PR #213, opened 2024-02-22
Fix incorrect native asset URL in README

BladeRunner713:
## Type of change

- Documentation

## Changes

- Native Asset link in README


bitzoic (approved):
Thanks for fixing this!

PR #224, opened 2024-03-27
Remove Workspaces

### LINKED ISSUE #141: Consider moving away from using workspaces
### Issue
The sway-libs repository currently uses workspaces to organize each library. This separates each library into its own project and allows for each library to be imported using only the library name. This does however comes with a major downside, being that each library must be imported individually. 

For example, here we would like to use 3 different libraries; StorageString, String, and Ownership. Adding these as dependencies would look like the following:

```rust
[dependencies]
string = { git = "https://github.com/FuelLabs/sway-libs", tag = "v0.8.1" }
ownership = { git = "https://github.com/FuelLabs/sway-libs", tag = "v0.8.1" }
storage_string = { git = "https://github.com/FuelLabs/sway-libs", tag = "v0.8.1" }
```

Our Sway contract would then import these as so:
```rust 
contract;

use string::String;
use storage_string::StorageString;
use ownership::owner;
```

### Proposed Change

Instead, if we organize the repository akin to the std lib with a root `Forc.toml` file, we would have the following dependencies and imports:

```rust
[dependencies]
sway_libs = { git = "https://github.com/FuelLabs/sway-libs", tag = "v0.8.1" }
```

```rust 
contract;

use sway_libs::{string::String, storage_string::StorageString, ownership::owner};
```

## Motivation

As more libraries are introduced, it may make sense to move away from workspaces to reduce the number of individual imports.

bitzoic:
## Type of change

## Changes

- All imports have changed from multiple imports to a single imports using the following:
```sway
libraries = { git = "https://github.com/FuelLabs/sway-libs", tag = "v0.4.0" }
```
From there your desired library may be imported:
```sway
use libraries::ownership::only_owner;
```

- Many of the primary functions have been moved out of the library's folder. The documentation for this will be moved into a book in a follow up PR.
- There are a number of private modules that are now exposed until https://github.com/FuelLabs/sway/issues/5765 is resolved.

## Notes

- This should get `forc doc` to produce a single document containing all libraries rather than multiple documents for each library.

## Related Issues

Closes #141


K1-R1 (dismissed):
Changes are good. I'd prefer the project name to be more specific, maybe just `sway_libs`?

SwayStar123 (approved):

K1-R1 (approved):

PR #240, opened 2024-05-01
Fix unused lines from fixed point library

### LINKED ISSUE #233: Reduce Warnings when building Libs
### Motivation

There are a number of warnings that appear when importing Sway Libs. These should be reduced as much as possible for Sway-Libs users.

SwayStar123:
## Type of change

## Changes

Removed all the unused lines from the fixed point library

Closes #233

### DIFF
--- a/libs/src/fixed_point/ifp128.sw
+++ b/libs/src/fixed_point/ifp128.sw
@@ -447,7 +447,6 @@ impl IFP128 {
     /// ```
     pub fn round(self) -> Self {
         let mut underlying = self.underlying;
-        let mut non_negative = self.non_negative;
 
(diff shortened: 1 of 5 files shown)


K1-R1 (approved):

SwayStar123:
Some of the removed lines are redeclaring an existing variable, so im gonna investigate if theyre actually unnecessary or just accidentally have an extra `let`. (Tests pass so might need to improve those if the latter is the case)

SwayStar123:
As far as i can tell the unused lines are just unused, and are not actually necessary. Probably left in accidentally while copy pasting

K1-R1 (approved):

PR #308, opened 2024-11-18
Remove comments on cross-contract reentrancy venerability

### LINKED ISSUE #307: Cross Contract Reentrancy Guard
### Motivation

Using manual storage slots, a cross contract reentrancy guard should be implemented. It should be noted and well documented that it is more limited than the existing reentrancy guard and use should be extensively documented.

bitzoic:
## Type of change

- Documentation

## Changes

- Removed cautionary note on cross-contract vulnerability in inline docs
- Removed comment on cross-contract vulnerability in documentation
- Added note stating cross-contract reentrancy not possible but to still exercise caution

## Notes

- Cross-contract reentrancy occurs when a contract like a vault issues and manages assets for token contract. However, as Fuel uses native assets, no contract call must be made to update balances. Therefore it is not possible to perform a cross-contract reentrancy attack. 
- A cautionary note on relying on other contracts for state has been added as this can introduce dependency attacks.

## Related Issues

Closes #307 

## Checklist

- [x] I have linked to any relevant issues.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have updated the documentation where relevant (API docs, the reference, and the Sway book).
   - [x] If my change requires substantial documentation changes, I have [requested support from the DevRel team](https://github.com/FuelLabs/devrel-requests/issues/new/choose)
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] I have added (or requested a maintainer to add) the necessary `Breaking*` or `New Feature` labels where relevant.
- [x] I have done my best to ensure that my PR adheres to [the Fuel Labs Code Review Standards](https://github.com/FuelLabs/rfcs/blob/master/text/code-standards/external-contributors.md).
- [x] I have requested a review from the relevant team or maintainers.
- [x] I have updated the changelog to reflect the changes on this PR.

### DIFF
--- a/docs/book/src/reentrancy/index.md
+++ b/docs/book/src/reentrancy/index.md
@@ -2,16 +2,11 @@
 
 The Reentrancy Guard Library provides an API to check for and disallow reentrancy on a contract. A reentrancy attack happens when a function is externally invoked during its execution, allowing it to be run multiple times in a single transaction.
 
-The reentrancy check is used to check if a contract ID has been called more than
-once in the current call stack.
+The reentrancy check is used to check if a contract ID has been called more than once in the current call stack.
 
 A reentrancy, or "recursive call" attack can cause some functions to behave in unexpected ways. This can be prevented by asserting a contract has not yet been called in the current transaction. An example can be found [here](https://swcregistry.io/docs/SWC-107).
 
 For implementation details on the Reentrancy Guard Library please see the [Sway Libs Docs](https://fuellabs.github.io/sway-libs/master/sway_libs/reentrancy/index.html).
-
-## Known Issues
-
-While this can protect against both single-function reentrancy and cross-function reentrancy attacks, it WILL NOT PREVENT a cross-contract reentrancy attack.
 
 ## Importing the Reentrancy Guard Library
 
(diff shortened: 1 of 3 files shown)


docs/book/src/reentrancy/index.md
```diff
@@ -45,3 +40,7 @@ To check if the current caller is a reentrant, you may call the `is_reentrant()`
 ```sway
 {{#include ../../../../examples/reentrancy/src/main.sw:is_reentrant}}
 ```
+
+## Cross Contract Reentrancy
+
+Cross-Contract Reentrancy is not possible on Fuel due to the use of Native Assets. As such, no contract calls are performed when assets are transferred. However standard security practices when relying on other contracts for state should still be applied, especially when making external calls.
```
SwayStar123, line 46: Isnt reentrancy a catchall term that includes contract calls too? I believe the contract call triggered from a payment on ethereum is just a special case that is particularly emphasized, but its not the only way for reentrancy to occur. Correct me if im wrong.

SwayStar123 (changes requested):
I thought you added tests to prove that cross contract reentrancy is stopped by the guard, i dont see it here

bitzoic:
> I thought you added tests to prove that cross contract reentrancy is stopped by the guard, i dont see it here

These tests already exist https://github.com/FuelLabs/sway-libs/blob/03f8ff9939471f0a5b6659e1b91c14e4bf39a9b7/tests/src/reentrancy/reentrancy_target_contract/src/main.sw#L58

docs/book/src/reentrancy/index.md
bitzoic: Yes it occurs anytime a function in a contract is called again. This requires either a second call within the contract(intra-contract) or if an external contract call is made which then re-enters the contract again. This often happens on Ethereum when payments are made as code is executed when this happens, meaning the contract could be malicious. But since no external calls are made, only when deliberately calling a contract, no accidental external calls can occur.

SwayStar123 (approved):

K1-R1 (approved):

